PR #33, opened 2017-10-24
Group System

meyer9:
None

Add a group system in preparation for contracts and shared career mode.

This PR isn't finished quite yet, but I've started adding some of the groundwork.

- [ ] Add `leave` network message
    - [ ] Leaving as owner should disband group.
- [ ] Add `transferOwnership` network message
- [ ] Add GUI

Also, I've allowed edit from maintainers, so you can commit to my branch if you want to work on this. Thanks!

### DIFF
--- a/Client/Client.csproj
+++ b/Client/Client.csproj
@@ -72,6 +72,9 @@
     <Compile Include="Base\Interface\IWindow.cs" />
     <Compile Include="LunaLog.cs" />
     <Compile Include="Base\MessageSystem.cs" />
+    <Compile Include="Systems\Groups\GroupMessageHandler.cs" />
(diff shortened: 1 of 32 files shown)


meyer9:
I get an error when syncing for some reason. I debugged it and it seems to have something to do with deserializing client->server group messages. Not sure, but I'll try to figure it out a bit later.

DaggerES:
Hi! Really nice work! anything else you need just let me know, the performance of LMP should be now much better as how it was in July. Also I'm now working on updating the orbit between the vessels, once that's finished I want to fix the warp system and also syncing the vessel changes (solar panels extend and other events) if you need some help or have some questions let me know trough here or skype (gabriel.spain)

PR #98, opened 2018-03-15
Build a KSPManager.exe for Luna

4669842:
Simple script to manage Luna updates and KSP instances.

- Has fallback to ensure directories not in use before updating.

- Directs users to updater if not found in the directory, also checks if .bat file is in the correct directory before installing.

- Has log generation in case failure to update.
**
Designed for Windows operating systems. Linux can use something more fluid. The BAT would work better if you built a manager.exe which follows the script.**

### Fixes included in this PR:  Dealing with Luna updates manually. People can boot from this to be easily upto date.

If a build is designed, it needs to be more gentle during the update part to ensure player data isn't removed.

### DIFF
--- a/Scripts/LunaManager.bat
+++ b/Scripts/LunaManager.bat
@@ -61,8 +61,8 @@ ECHO 2.  Play Kerbal Space Program
 ECHO Q.	 Quit the Manager
 ECHO -------------------------------------
 ECHO  Invalid choices will default to installing
-ECHO  the client files as I want to keep autonomous. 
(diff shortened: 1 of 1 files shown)


4669842:
Maybe, I was thinking it should be an independent build that works with your updater.exe

If you could merge it into the updater.exe itself it would remove the need to have a second .exe created.

gavazquez:
I agree, but the Updater.exe and all that code is in the other repo

PR #102, opened 2018-03-16
Fixed server command /listclients always says "No Clients connected" …

Sladernimo:
…doesn't matter if there are clients connected or not.


gavazquez:
Thanks!!

PR #117, opened 2018-03-22
Share Player Progress

MalteJanz:
Added an experimental system for sharing the progress between players.
Currently the following data is in sync:
-Funds
-Science
-Reputation
-Tech tree unlocks
-Contracts

Reverting is at the moment not supported (it will cause some loses / winnings or unexpected bugs).
To use this system set the "GameMode" setting to "Science" or "Career" and also the the "ShareProgress" setting to true.

The system is build under the following concept:
Do the things that happen on one client also on the other clients. That means if one client accepts an contract, the other client should beheave the same as he manually accepted that contract too. It is also mostly event driven and no big savegames (/ scenarios) are carried over the network.

Keep in mind that this system is in a very early stage and there will be many bugs that needs to be discovered and fixed.

In the future it would be nice if this system work together with the group system so there can be different groups of players with the same progress and there would be a little competition.


Zer0t3ch:
@MalteJanz Are there any issues with using subspace warp type with shared progress? I seem to recall seeing something about that somewhere, but I can't find it any more.

PR #148, opened 2018-06-06
Removes space in log filename

AlexFaraino:
none
Removes space in log filename, because it's a pain to handle on a non-GUI system ;)


gavazquez:
You're totally right ;)

AlexFaraino:
Who ! That was quick 👍

PR #164, opened 2018-06-20
Added some more fields to ModuleWheelMotor.xml

Alfred-PictoCube:


gavazquez:
Just one thing. If I for example modify the wheel motor while in flight and set to inverted and leave the vessel this means that if another player takes it it won't have that fix. Are you sure we shouldn't sync that?

Alfred-PictoCube:
I'd be more surprised having wheels set to backward whereas that's not an expected behavior.
The evil is in the details, but if you think there are some real purpose on it, that's your decision :P

gavazquez:
You're probably right

Issue #206, opened 2018-10-08
Possible unintended bypass of lifetime check of underlying Unity engine object


thygrrr:
A lot of checks in the code are done with the null propagating operator, ?.

These checks are not good when used with anything that inherits from UnityEngine.Object, because the managed portion of the object may be non-null, but its native equivalent may be.

So Unity Technologies actually went ahead and... **GASP** overloaded the equality and inequality operators, and added a boolean cast operator to UnityEngine.Object (and thus all its descendants, including GameObject, Transform, MonoBehaviour, ...)

In particular, an object may be in the process of being Destroy()'ed, which is when this boolean cast would return "false" on a technically non-null C# object.

I've created a PR with this change -> #207

thygrrr:
Do you think the major singletons need checking? 

KSCVesselMarkers.fetch, Planetarium.fetch, etc.

The original Squad code usually doesn't, as they are initialized in Awake(), but of course if they aren't "DontDestroyOnLoad", that means they would get re-initialized with certain scene changes.

PR #207, opened 2018-10-08
Possible unintended bypass of lifetime check of underlying Unity object

thygrrr:
Quoting #206 

Further reading: https://github.com/JetBrains/resharper-unity/wiki/Possible-unintended-bypass-of-lifetime-check-of-underlying-Unity-engine-object

A lot of null checks were done with the null propagating ?. operator, regardless of the underlying type. Using Jetbrains Rider's Unity Plugin helped me identify these occurrences; and even though not all of them were in crucial sections, especially when gameobjects could get destroyed, better safe than sorry.

These checks are not good when used with anything that inherits from UnityEngine.Object, because the managed portion of the object may be non-null, but its native equivalent may be.

So Unity Technologies actually went ahead and... ***GASP*** overloaded the equality and inequality operators, and added a boolean cast operator to UnityEngine.Object (and thus all its descendants, including GameObject, Transform, MonoBehaviour, ...)

In particular, an object may be in the process of being Destroy()'ed, which is when this boolean cast would return "false" on a technically non-null C# object.

The boolean null check is the same that underlies the == and != operator overloads.

- using the boolean style null check in most places
- we could use the != null check instead, which feels much cleaner but adds even more clutter
- we should NOT use the null coalescing operator or null propagating operator on descendants of UnityEngine.Object
- setting up a warning like the one in Rider would greatly help avoid this
- YES this is my **least, very least favourite** "feature" or "design flaw" of Unity. It doesn't seem to be on their deprecation list at all.

### Please review:
Two pairs of eyes are better than one in this case. I'm afraid I might have inadvertently reversed some logic through my changes. (i.e. missed a ! or set some parentheses wrong for a new &&)


gavazquez:
Damn... that leaves the code much more ugly :(
I totally agree with you tough and it seems that it's necessary to do what you're doing....

I will have a look at your changes and implement the same wherever I can

thygrrr:
It's even worse. 
https://blogs.unity3d.com/cn/2014/05/16/custom-operator-should-we-keep-it/

> The custom ==operator is not thread safe, so you cannot compare objects off the main thread. (...)

Not sure if they ever fixed that.

gavazquez:
Oh come on... They really need to tackle this stuff, specially since multithreading is the future.
A lot of stuff needs to be changed on LMP :/

thygrrr:
I don't think LMP needs THAT many changes, everything that IS read or written in another thread and is a UnityEngine.Object was a danger to the main thread code anyway.

I am having a hard time understanding what runs in what thread with LMP, still.

gavazquez:
IIRC nothing that is from unity is handled in another thread. I took care of that as unity complains a lot when something is touched/read in another thread.
Also, on client side, only the network system is actively running in another thread, everything else is in the main thread

thygrrr:
https://forum.unity.com/threads/unity-2017-and-null-conditional-operators.489176/

This has some discussions and some solutions people have experimented with.

thygrrr:
Last one, for good laughs... https://github.com/JetBrains/resharper-unity/issues/148

Because on ~~Crystal Meth~~ Unity, this != null can be false.

PR #303, opened 2019-07-19
Big Craft Welcome

awabast:
<?xml version="1.0" encoding="utf-16"?>
<DedicatedServerSettingsDefinition xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
  
  <UseRainbowEffect>true</UseRainbowEffect>
  
  <Red>255</Red>
  
  <Green>100</Green>
  
  <Blue>10</Blue>
</DedicatedServerSettingsDefinition>

MasterServersList/DedicatedServersList.txt


awabast:
I really would like my server big craft welcome to be dedicated

Issue #328, opened 2019-12-30
Kerbals go MIA when recovering a vessel


twiclo:
Remember to attach the "KSP.log" and the "output_log.txt" whenever is possible!  
You can find the KSP.log in ".\Kerbal Space Program\KSP.log"  
And output_log.txt in "C:\Users\YOURUSERNAME\AppData\LocalLow\Squad\Kerbal Space Program\output_log.txt"
(you probably need to display hidden folders to get into AppData)

In case you are using a nighly build of LMP from appveyor include the build number.
For example: 0.0.9.**384**
Otherwise, specify the version of LMP you are using from the github releases.

**If you have access to the server. Please zip the whole "Universe" and "Config" folder.
Also if you are running mods in the server, attach the "LMPModControl.txt"**

------------------------------------------------------------------------------
------------------------------------------------------------------------------
------------------------------------------------------------------------------

OS: Fedora 30

------------------------------------------------------------------------------

KSPVersion: 1.8.1.2694 (LinuxPlayer)

------------------------------------------------------------------------------

LMP version (include the build number if you are using a nighly build): 0.24.0

------------------------------------------------------------------------------

MODS (add link to download them): Planetary base systems: https://www.curseforge.com/kerbal/ksp-mods/kerbal-planetary-base-systems

------------------------------------------------------------------------------

Expected behaviour: Kerbals should not be marked MIA

------------------------------------------------------------------------------

Current behaviour: Kerbals will be marked as mia when recovering a vessel. It doesn't seem to matter if it's from the launch pad or from a flight.

------------------------------------------------------------------------------

Steps to reproduce: We haven't been able to consistently reproduce this. It seems to happen about 50 percent of the time

gavazquez:
Let me know if it still happens. I uploaded a fix

PR #345, opened 2020-05-09
Add Per-Player Kerbals

Erdroy:
No fixes included.

Allows Players to have their own Kerbals and thus, removes the need to wait for locks (if a Player want to use the same Kerbal). This may fix some issues with the syncing (e.g. #328 ).

This creates a directory using Player's UID (might be changed to just PlayerName), Universe/PlayerData/PLAYER_UID/
Where player's Kerbal data is stored. This could be extended to allow more per-player data, and probably will be required to implement groups/companies etc.


Erdroy:
Well, actually this is not a good option. After rethinking everything - as credits etc. are sharedm this is not a good option. I want to just jump into making companies instead.

I'll close this draft now, sorry ;)

gavazquez:
Hi! thanks for the PR anyway and if you need some with the code just text me on discord

PR #347, opened 2020-05-28
Fix reverting not restoring funds

lgoenner:
Reverting to vehicle assembly currently does not return funds


gavazquez:
Thanks mate!

PR #349, opened 2020-05-30
Killing contracts is not enough

lgoenner:
I have encountered countless rescue missions even they should be disabled, so I did some digging.
It seems calling `Kill()` on a contract does not force the game to instantly remove it (documentation uses the word [prepares](https://kerbalspaceprogram.com/api/class_contracts_1_1_contract.html#a423e9e110fc485183a48865a926c7665
)) `Withdraw()` seems to correctly prevent the contracts from appearing.

Withdrawing contracts in the `!LockSystem.LockQuery.ContractLockBelongsToPlayer(SettingsSystem.CurrentSettings.PlayerName)` block makes available contracts invisible for all other players except the one holding the contract lock. Although this seems to be the intention of the code, it would be a regression, as all players can currently see and accept the same available contracts. What should be done about this?


gavazquez:
The contract lock is used so **only 1 player** generate contracts in the whole game. You're right about the withdraw btw ;)

PR #448, opened 2021-07-11
Expand IPv6 support; Attempt IPv6 connections via server list

DasSkelett:
Follow-up to #445, this basically "completes" IPv6 support.
This PR extends the IPv6 support to server-list connections as well, with firewall punchthrough support.
It touches a lot of code, it changes a lot of the core networking logic. Please review carefully.

This PR also includes changes to the wire format of some messages. I.e. master servers / servers / clients receiving from older or newer master servers / servers / clients will probably throw some NullReferenceExceptions and have other unexpected behavior. Do we need to bump some version number for this, or is this just the danger of running nightlies?

## Testing

This has been tested
- with IPv4-only server & master server running on localhost
- with dual-stack server & master server running on localhost
- with IPv4-only server running on remote machine A and master server running on remote machine B:
  - with "static port mapping" NAT & firewall between client <-> server/master server and firewall between server <-> master server
  - with "symmetric" IPv4 & firewall between client <-> server/master server and firewall between server <-> master server
- with dual-stack server running on remote machine A and master server running on remote machine B

It has not been tested with an "IPv4-only" client yet, as I need to start a VM for that, since I rely on IPv6 on my main machine.

I have some Docker images on Docker Hub, if it helps with testing: [Master server](https://hub.docker.com/r/dasskelett/lmpmasterserver) | [Server](https://hub.docker.com/r/dasskelett/lmpserver).
However they can be changed at any time, and do not necessarily match the code currently pushed to this branch.
For example, they don't support UPnP, as that code fails to compile on my machine (something with the Visual Studio DLLs) and is commented out in my working directory.

I have a master server and a server using the above Docker images running on a VPS, whose addresses I can share for testing if you are interested.

### Client
- Initiate single global ClientConnection at startup, don't replace after disconnect
- Make socket IPv4-only if OS doesn't support IPv6
- Implement IPv6->IPv4 fallback logic in ConnectToServer for direct connections
- The manual retries in `ConnectToServer()` have been removed and replaced by Lidgren's builtin retry logic configured via `Config.MaximumHandshakeAttempts` and `Config.ResendHandshakeInterval`
- `SendNetworkMessage()` starts the `NetClient` when sending unconnected messages to the master servers if necessary, usually when retrieving the server list.
- Servers in the server list are now updated when receiving new data (especially useful for debugging, but might benefit other situations as well) 
- Communicate IPv6 addresses with master servers when asking for NAT introduction
- `ServerIsInLocalLan()` compares the prefix for IPv6 addresses (thanks to the global uniqueness of GlobalUnicastAddresses this is fairly reliable)
- Add debug option to overwrite the master server list
- Show IPv6 RTT in server list

### Server
- Communicate IPv6 addresses with master servers
- Add debug setting to overwrite master server list
- Allow specifying a separate `ListenAddress` for the web server, with fallback to the game server `ListenAddress` if left unspecified.

### Master server
- Support IPv6 addresses for servers, send to clients
- Send another NAT introduction for IPv6, if both client and server have IPv6 enabled:
  - The first NAT introduction message triggers hole punching for both the public IPv4 addresses and the IPv6 addresses. This is unavoidable, since we need to send the trigger message to the public IPv4 addresses of the client and server.
    This one is likely to succeed in the most common case (as long as both client and server have working IPv6, or the IPv4 NAT in between allows hole punching).
  - The second NAT introduction message triggers hole punching for both the public IPv4 addresses again and the internal IPv4 addresses. It is sent after a delay of 50 ms, to allow some cooldown after the first round (even though my testing showed there's no problem with sending dozens of NAT punch-through messages, the connection setup after a punch-through succeeded is race-free).
  - Since we have this delay of 50ms, the whole operation now happens in a background thread, as to not block receiving and processing other messages in the meantime.
- Fail softly if it can't get its own external IPv4 address, e.g. if networking of the container is broken \*ahem\*
- Show IPv6 address in HTTP interface table
- Add debug option to overwrite the master server list

### Common
- `CreateAddressFromString()`  now supports parsing IPv6 addresses, however it does still filter out AAAA records of hostnames. This is not a problem, because it is only used to read the master and dedicated server lists, both of which must run on IPv4 (or dual-stack) as of now. At some point it should be refactored to return an array of addresses, including IPv6.
- `CreateEndpointFromString()` returns all addresses for a hostname, instead of the first. This fixes the direct connection feature if a hostname has an AAAA record but the server (or client) are not IPv6-enabled.  This was already broken before #445.
- A new `GetOwnInternalIPv6Address()` can return any available GUA address, it filters out link-locals and ULAs, first of which are non-routable and the second indicating that the machine is behind IPv6 NAT (ULA's are roughly equivalent to IPv4's RFC1918 subnets, 10.0.0.0/8 etc.).

### Build
- Some of the `PostBuildEvent` statements have been fixed to not run on *nix systems, they don't have `xcopy`.
- The DLL references in `MasterServer.csproj` are updated to match the ones in `LmpMasterServer.csproj`, so they are copied to the output directory, and it doesn't crash at startup trying to import the DLL.

## Future enhancements
- Look into port opening for IPv6 using [PCP](https://en.wikipedia.org/wiki/Port_Control_Protocol). It's not as heavily needed for the server, as firewall punch-through is much more likely to succeed via IPv6 due to the lack of NAT. And for the master server you should set up manual port forwardings/openings anyway for stability.
- Look into better ways to determine whether a server is on the same network as the client for IPv6 addresses. Maybe by checking whether the route has a gateway or is on-link?
- Make servers listen on IPv6 by default after thorough testing, so even casual users who don't look too deep into config files experience the benefits of IPv6.
- Support IPv6-only servers, make master servers listen on IPv6 as well. There is probably not much demand for this, and it is a little trickier to do, but maybe it's useful for some people.

### DIFF
--- a/LmpClient/Localization/Structures/ServerListFiltersText.cs
+++ b/LmpClient/Localization/Structures/ServerListFiltersText.cs
@@ -4,7 +4,8 @@
     {
         public string Title { get; set; } = "Server list";
         public string NoServers { get; set; } = "No servers!";
-        public string Ping { get; set; } = "Ping";
+        public string Ping { get; set; } = "IPv4 Ping";
+        public string Ping6 { get; set; } = "IPv6 Ping";
         public string Players { get; set; } = "Players";
         public string MaxPlayers { get; set; } = "Max players";
         public string Mode { get; set; } = "Mode";
--- a/LmpClient/Network/NetworkConnection.cs
+++ b/LmpClient/Network/NetworkConnection.cs
@@ -2,13 +2,13 @@
 using LmpClient.Base;
 using LmpClient.ModuleStore.Patching;
 using LmpClient.Systems.Network;
-using LmpClient.Systems.SettingsSys;
 using LmpCommon;
 using LmpCommon.Enums;
 using LmpCommon.Message.Base;
 using System;
 using System.Net;
 using System.Threading;
+using UniLinq;
 
 namespace LmpClient.Network
 {
@@ -50,20 +50,20 @@ namespace LmpClient.Network
             }
         }
 
-        public static void ConnectToServer(string address, int port, string password)
+        public static void ConnectToServer(string hostname, int port, string password)
         {
-            ConnectToServer(new IPEndPoint(LunaNetUtils.CreateAddressFromString(address), port), password);
+            ConnectToServer(
+                LunaNetUtils.CreateAddressFromString(hostname).Select(ep => new IPEndPoint(ep, port)).ToArray(),
+                password
+            );
         }
 
-        public static void ConnectToServer(IPEndPoint endpoint, string password)
+        public static void ConnectToServer(IPEndPoint[] endpoints, string password)
         {
-            if (MainSystem.NetworkState > ClientState.Disconnected || endpoint == null)
+            if (MainSystem.NetworkState > ClientState.Disconnected || endpoints == null || endpoints.Length == 0)
                 return;
 
             MainSystem.NetworkState = ClientState.Connecting;
-
-            if (NetworkMain.ClientConnection.Status == NetPeerStatus.NotRunning)
-                NetworkMain.ClientConnection.Start();
 
             SystemBase.TaskFactory.StartNew(() =>
             {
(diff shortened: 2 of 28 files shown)


DasSkelett:
Whoops, looks like the server has a "CPU leak", it completely hogs one CPU. Gonna have to check whether that's caused by one of my changes, or has been present before.
image

DasSkelett:
@gavazquez I have a `UnityPlayer.so` for debugging on Linux, would you be interested in adding it to the repository, next to the `UnityPlayer.dll` for Windows?

gavazquez:
> 
> 
> @gavazquez I have a `UnityPlayer.so` for debugging on Linux, would you be interested in adding it to the repository, next to the `UnityPlayer.dll` for Windows?

Definatelly! Feel free to add it and update the wiki if you can:
https://github.com/LunaMultiplayer/LunaMultiplayer/wiki/Debugging-in-Visual-studio

PR #452, opened 2021-07-16

BPike-KueMedia:


DasSkelett:
```
games.kuemedia.com.au:8800 (Sandbox)
games.kuemedia.com.au:8801 (Modded Sandbox)
```
I don't think the parsers can handle arbitrary strings after the hostname:port pair, at least I'm intercepting exceptions with my debugger when it tries to resolve these two hosts.

Also it looks like these two servers are listening on different ports than what has been specified in the dedicated server list, http://servers.lunamultiplayer.com:8701/ shows `119.18.17.10:51564` and `119.18.17.10:18049`.

I'm gonna do a commit to fix the syntax of the file, but I'll keep the port numbers for now.
@BPike-KueMedia, if you want your servers to show up as dedicated, you need to make sure they're also listening on these ports _to the outside_, i.e. NAT doesn't mangle these ports.

gavazquez:
Shit this happened as I did the merge on the phone and didn't checked the file :/

PR #453, opened 2021-07-17
Fix master server auto-update with /nightly flag

DasSkelett:
(A version number is assumed to be in the form of MAJOR.MINOR.PATCH.BUILD in the following text, with BUILD being optional in most cases)

## Problems
The master server auto-update functionality for nightly releases is currently broken.
- It tries to reach the AppVeyor API using `RepoConstants.ApiLatestGithubReleaseUrl`, which should be `RepoConstants.AppveyorUrl` instead (nightly releases aren't published to GitHub)
- Even with the URL fixed, it'll only ever run the update when either MAJOR, MINOR or PATCH are bumped. This doesn't seem to happen very often. For nightly builds, only the BUILD number is incremented.

As an example, none of the "nightly" master servers seems to have picked up my IPv6 changes yet (which is easy to spot with the missing "IPv6 address" column in the HTTP UI).

- The API URL is fixed
- The auto-update logic also compares the BUILD parts of the DLL and the latest nightly build, basically making the nightly master servers update themselves after every commit to master (with a delay, of course).
- Local builds don't have a BUILD number in the assembly version, which we now use as indicator _not_ to run any updates (so it doesn't mess with debugging).
  This also prevents e.g. master servers running in Docker containers from doing auto-updates, but _usually_ this is desired. We could think about a `/force-update` flag or similar if there's demand to keep auto-updates in containers or other self-compiled deployments. Not sure how master server owners are running them, input welcome.
  
A manual update is required for these changes to reach already deployed master servers and have an effect.

### DIFF
--- a/LmpUpdater/Appveyor/AppveyorUpdateChecker.cs
+++ b/LmpUpdater/Appveyor/AppveyorUpdateChecker.cs
@@ -19,7 +19,7 @@ namespace LmpUpdater.Appveyor
                     {
                         wc.Headers.Add("user-agent", "Mozilla/4.0 (compatible; MSIE 6.0; Windows NT 5.2; .NET CLR 1.0.3705;)");
 
-                        var json = wc.DownloadString(RepoConstants.ApiLatestGithubReleaseUrl);
+                        var json = wc.DownloadString(RepoConstants.AppveyorUrl);
                         return JsonConvert.DeserializeObject<RootObject>(JObject.Parse(json).ToString());
                     }
                 }


gavazquez:
Hi!. I totally agree on the problem described here:

_It tries to reach the AppVeyor API using RepoConstants.ApiLatestGithubReleaseUrl, which should be RepoConstants.AppveyorUrl instead (nightly releases aren't published to GitHub)_

Regarding the build number, I'm not so comfortable ont riggering an update of the nightly master server on every new build since the way it starts and restarts is VERY fragile and it fails quite a lot of times.

Instead, what we should do is increase the version number for nightly, this way we will force the nightly master servers to update themselves.

Here is a wiki that more or less explains the versioning system we're using:

https://github.com/LunaMultiplayer/LunaMultiplayer/wiki/Versioning-format

If you want I can fix this PR so it only fixes the first point and then update the version number, or as you prefer

DasSkelett:
Okay, sounds good. I'm gonna revert everything but the API URL change then. I'll leave the version number update to you, as I don't know about all the places where it needs to be changed.

gavazquez:
Great. here is the commit for increasing the version num:
https://github.com/LunaMultiplayer/LunaMultiplayer/commit/c485e8588fd7748adee8dc66f055d398bbcdc618

PR #461, opened 2021-08-05
Update list - adds ABrit's second server

ABritInSpace:
Add ABrit's second server (modpack).


DasSkelett:
That server is password protected. I'm not sure if @gavazquez has a policy for dedicated servers, but I don't think adding a password-protected server to the dedicated list makes any sense.

ABritInSpace:
Good point, probably shouldve thought this through at the time. Thanks!

PR #462, opened 2021-08-06

rf152:
Added server


DasSkelett:
That server is password protected. I'm not sure if @gavazquez has a policy for dedicated servers, but I don't think adding a password-protected server to the dedicated list makes any sense.

PR #465, opened 2021-08-09
move LMPModControl.xml to Config directory

### LINKED ISSUE #463: move LMPModControl.xml
Have the `LMPModControl.xml` file be in the Configs folder rather than in the root dir of the server

zivoy:
Moving the LMPModControl.xml to the Config folder. It makes more sense to have it there, and it makes deploying docker containers easier since you won't have to make a volume just for the file if you are playing with mods.

This resolves #463 (it was a more simple change than I thought)


DasSkelett:
This would break servers when they upgrade, the file would have to be moved manually. It would be nice if you could provide a migration logic, i.e. moving the old file from the server root to the Config directory.

zivoy:
this should be fine, but it will through an error if there is already a LMPModControl file in the Config folder. though I don't believe this would be an issue

PR #493, opened 2022-04-11
Add debug stage to server Dockerfile, rework master server Dockerfile

DasSkelett:
This reworks both Dockerfiles. The master server Dockerfile didn't actually built the project from source, but tried to download a build artifact from AppVeyor (which is broken).

## Changes
* Actually build master server from source
* Split into multiple stages and steps to better make use of build caching/layering
* Add debug stage to server Dockerfile, which sets some env vars to enhance dotnet debugging, and has `/bin/ash` as CMD so one can attach to the container and start the server using `dotnet run`

The images aren't exactly minimal, especially the master server one (~700MB). But mono:slim isn't enough to run the server, unfortunately. And mono's `mkbundle` to generate a self-contained executable is not very usable.

Based on this one could set up CI to automatically build and push the images to a Docker Registry of ones choice, e.g. [ghcr.io](https://github.com/features/packages).

I've had this sitting in a branch for a long long time, but now that I'm running my own master servers I think it's worth sharing and preserving the files here.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,7 @@
+**/bin/
+**/obj/
+.git/
+/Documentation/
+
+/Dockerfile*
+/External/KSPLibraries/KSPLibraries.7z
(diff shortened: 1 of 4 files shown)


gavazquez:
That's amazing!
Thanks a lot for sharing it!

PR #513, opened 2022-10-29
Add new master server

xBelladonna:
None

I'm running a public master server and am adding it to the master servers list.

List: https://lmp.nightshade.fun/ or http://lmp.nightshade.fun:8701/
Server: lmp.nightshade.fun:8700

### DIFF
--- a/MasterServersList/MasterServersList.txt
+++ b/MasterServersList/MasterServersList.txt
@@ -20,3 +20,5 @@ direct.imexile.moe:8702
 ms.lmp.dasskelett.dev:8700
 #DasSkelett Nightly. List: https://ms-nightly.lmp.dasskelett.dev or http://116.203.125.175:8751/
 ms-nightly.lmp.dasskelett.dev:8750
+#Nightshade. List: https://lmp.nightshade.fun or http://lmp.nightshade.fun:8701/
+lmp.nightshade.fun:8700


DasSkelett:
Very nice, thanks @xBelladonna! Having one in Australia is handy for all the people on the other side of the planet, as the current ones are mostly in Europe or the US.

There's one issue with your domain though: when querying for the A record, it returns one directly, pointing to your home IP address (I suppose).
But when querying for any other record type, e.g. AAAA, NS or TXT, it returns a CNAME to the parent domain (`nightshade.fun.`). This isn't allowed in the DNS standards, a CNAME is only allowed to exist alone by itself, with no other records, on a specific subdomain. This can lead to problems with DNS resolvers rejecting the domain and returning SERVFAIL for example.
And the A and AAAA records of the base domain also point to Cloudflare. This is a problem, because Cloudflare doesn't proxy any traffic besides HTTP and HTTPS on port 80 and 443. LMP's UDP game traffic to port 8700 is dropped.
So depending on which path somebody's resolver happens to take, the CNAME or the direct A record, their server may or may not be able to reach your master server.

I suggest to remove the CNAME for the `lmp.nightshade.fun` and only keep the direct A record.

Just out of interest, how's your setup? Are you running it "bare-metal", or in (Docker?) containers? If so, are you using our images from https://github.com/LunaMultiplayer/LunaMultiplayer/pkgs/container/lunamultiplayer%2Fmasterserver ?

xBelladonna:
Thank you for all that, it was incredibly informative. I've had a look into why this is happening and I can confirm it is not necessarily how my DNS is configured, but how CloudFlare itself is behaving when queried for my records. I do not actually have any A records for `lmp.nightshade.fun`, only one CNAME, which points to the A record for my base domain, which points to my server address. This one A record is proxied, hiding traffic behind CloudFlare's proxy, but all the CNAMEs are unproxied, exposing my IP address directly through the A record. I suppose this appears to be implemented by CloudFlare actually returning an A record when asked instead of the CNAME to my base domain and server IP as I've configured. I'm not sure how to really get around this apart from unproxying my base domain, which I honestly don't mind doing, so that's now fixed.

To answer your question about my server setup, I was actually compiling the master server and building a Docker image myself as that is the first I'm seeing of a location to even download it let alone a container of it, and have gladly switched over to using that. The server itself is running Proxmox hypervisor, with Docker behind a reverse proxy running inside a VM that's port forwarded to.

Hopefully this has sorted it all out. Thanks again for your informative reply.

DasSkelett:
Looks good now, DNS is cleaned up and a testing server could successfully reach the master server. Thanks!

> To answer your question about my server setup, I was actually compiling the master server and building a Docker image myself as that is the first I'm seeing of a location to even download it let alone a container of it, and have gladly switched over to using that. The server itself is running Proxmox hypervisor, with Docker behind a reverse proxy running inside a VM that's port forwarded to.

Yeah since there are only a handful of people dealing with master servers, we don't have that much documentation on running them. The official Docker images are also a relatively new addition.

Master server artifacts are also built as part of the CI flow, they can be found [here]( https://ci.appveyor.com/project/gavazquez/lunamultiplayer/branch/master) under "Configuration: Debug" -> "Artifacts" (e.g. https://ci.appveyor.com/project/gavazquez/lunamultiplayer/branch/master/job/7we7aom8gwk6tl25/artifacts). The artifacts are only retained for 30 days after build though.
But as long as you are using the Docker images or building your own this doesn't matter for you.

Thanks for the information, it's good to know what systems it runs on, especially for deciding what architectures to build the artifacts/images for.

DasSkelett:
https://github.com/LunaMultiplayer/LunaMultiplayer/wiki/Master-server-status is updated, I'll do the list in the README at some point later.
I've added you to the "nightly" list as you are using the current master Docker image (we weren't building the images yet at the last release, otherwise there would be a `:0.29` tag). As the current nightlies are still mostly compatible with the latest release, it doesn't really matter right now.
The wiki page is editable by everyone, so feel free to move it to the stable list if you think that's a better fit and plan to stay on a compatible version or track release tags once they are available.

PR #514, opened 2022-11-02
Update Nightshade master servers

xBelladonna:

In reference to #513, I have added a compiled release of 0.29.0 master server and also the nightly docker image from the ghcr. Both are live and I intend to keep both up and properly deploy both release and nightly once docker image tagging is properly tracked. I have already modified the wiki page to reflect these changes.

### DIFF
--- a/MasterServersList/MasterServersList.txt
+++ b/MasterServersList/MasterServersList.txt
@@ -22,3 +22,5 @@ ms.lmp.dasskelett.dev:8700
 ms-nightly.lmp.dasskelett.dev:8750
 #Nightshade. List: https://lmp.nightshade.fun or http://lmp.nightshade.fun:8701/
 lmp.nightshade.fun:8700
+#Nightshade Nightly. List: https://lmp.nightshade.fun/nightly or http://lmp.nightshade.fun:8751/
+lmp.nightshade.fun:8750


xBelladonna:
Also, I would very much love to see arm/arm64 images available for Docker if at all possible! I plan to run these master servers on my Raspberry Pi 4 and having images in those archs would be really useful for that.

DasSkelett:
Nice!
There seems to be some misconfiguration now, though. The new nightly master server already has a bunch of servers registered even though it isn't known to clients yet, and the existing "stable" master server only has relatively few servers compared to other master servers.

https://lmp.nightshade.fun/nightly/ :
nightly

https://lmp.nightshade.fun :
stable

Mine, https://ms.lmp.dasskelett.dev:
mine

Is it possible that you've accidentally put them in some sort of loadbalancing configuration? I.e. the new one is also listening on 8700 instead of 8750, and the packets are being sent to both of them in alternation?

> Also, I would very much love to see arm/arm64 images available for Docker if at all possible! I plan to run these master servers on my Raspberry Pi 4 and having images in those archs would be really useful for that.

For Docker we already have arm64 images, both for normal servers and for master servers.
If you go to https://github.com/LunaMultiplayer/LunaMultiplayer/pkgs/container/lunamultiplayer%2Fmasterserver/47959222?tag=master and click on the "OS/Arch" tab, you should see the list of current architectures.
(No need to pull them by their specific hash though, Docker should be able to figure out automatically which variant to pull based on the image manifest).

xBelladonna:
Ah, that... would probably be because I moved the existing nightly master server to the new nightly URL, compiled a release version of the master server and served that as the one our nightly master server used to handle. I did this because the docker image seems to just be tracking master and I can't find a release tag, so compiling it myself seemed to be the only option because appveyor didn't have an artifact that old. Is this okay or would I need to set this whole thing up another way so that the previously running container handles the nightly requests?

DasSkelett:
For those reading this some time in the future, we could resolve the problems by fixing a bug in the master server Dockerfile, which made dotnet build x64 binaries even for arm images.

And from @xBelladonna:
> If anyone has the same cursed Raspberry Pi setup with the arm64 kernel running armv7l userspace, you'll need to use the flags `--platform linux/arm64` and `--security-opt seccomp:unconfined` in order for the container to be able to execute anything, but it works!

PR #519, opened 2022-11-21
Add Barney's Server to dedicated server list

Barney295:


DasSkelett:
https://en.wikipedia.org/wiki/Private_network

DasSkelett:
This won't work. We need your public IPv4 address.

Please also have a read here: https://github.com/LunaMultiplayer/LunaMultiplayer/wiki/Dedicated-server
Your server is currently password protected, why is this?
Are you sure you want to mark your server as dedicated server?

Barney295:
Thanks for the quick feedback, I put a password on since I'm using Infernal Robotics. I changed it to public now. Can I send any config files or a screenshot of the console to help?

DasSkelett:
> I put a password on since I'm using Infernal Robotics.

I see. For modded servers we make exceptions, since you might want to tell people the mod list beforehand, but under the condition that a contact point is linked as website. You seem to be linking your Discord server there, so that would be fine. If you want you can set the password again, but please give it to everyone who asks after sharing the mod list, as dedicated servers are supposed to be public servers, accessible for everyone.

Now, your IP address doesn't appear to be static. First you've pushed `105.186.136.180:8800`, then `105.187.231.170:8800`, but now I see `105.184.178.179:8800` in the server list.
You either need a static IP address, or working (!) DynDNS to host a dedicated server, can you set it up and list the domain instead?

Barney295:
Yeah, sorry. I'm working on it; I thought I had it, but I have to work on some settings still

Barney295:
I'm unable to get a static IP. I think my ISP has restricted that, so I will close this. Sorry about the trouble; I appreciate your efforts

DasSkelett:
You can setup dynamic DNS if you still want to have your server listed as dedicated.
But in the end it might not be worth the trouble, the only advantage is the colour of the text in the server list ^^

PR #520, opened 2022-11-28
Updated my list of dedicated servers

ABritInSpace:
Added, respectively
 - Modpack Server
 - BDA Server
Both of which are public, with links to the discord (for mods)


DasSkelett (approved):
Looking good, thanks!

PR #524, opened 2022-12-11

ZariaGalaxy:


DasSkelett:
Is the IP address allocated statically to you? It appears to be from an ISP's pool for residential addresses, which tend to be dynamic.
It hasn't changed since you opened the PR, but nowadays they often (only) change on router reboot.
Can you confirm that this is indeed a static IP address?

PR #525, opened 2022-12-11

tetraflon:
Add KSPDogFightServer as dedicated server:
```
140.238.203.31:8001
140.238.203.31:19132
```

### DIFF
--- a/MasterServersList/DedicatedServersList.txt
+++ b/MasterServersList/DedicatedServersList.txt
@@ -28,3 +28,6 @@ checha.ru:8888
 92.221.152.39:8800
 #FireStryker
 ksp.lekekelparadox.xyz:8800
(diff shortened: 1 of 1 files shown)


DasSkelett:
Is the IP address allocated statically to you / are you running the servers on a server?
It hasn't changed since you opened the PR, but nowadays they often (only) change on router reboot.
Can you confirm that this is indeed a static IP address?

tetraflon:
> Is the IP address allocated statically to you / are you running the servers on a server? It hasn't changed since you opened the PR, but nowadays they often (only) change on router reboot. Can you confirm that this is indeed a static IP address?

yes the server is host on Oracle Cloud service so ip will not change unless oracle purged my public ip for some reason.

PR #529, opened 2023-01-06

Wimmetje1:


DasSkelett (approved):
Looks good, IP address / DynDNS was stable over multiple days.

PR #539, opened 2023-01-17
Lmp server fixes

JanePixel:

I have fixed two server bugs here:
- Custom flags did not display in-game
- Server would occasionally close and not restart when rebooted from the admin panel in the LMP menu window

Custom flags:
After some analysing it seemed that the client was expecting the field “NumBytes”. However, the server seems to not include this when sending the flag details. I have added the field and now custom flags seem to work again (tested with multiple players and several custom flags).

Server reboot by admin panel:
I have noticed several times that when the server is rebooted using the admin panel, it will have a small % chance of not rebooting and instead closing the console.
This seems to occur due to the fact that “QuitEvent.Set();” allows the “Main()” method to continue, occasionally reaching the end before “//Start new server” is reached, resulting in the application closing.
I have moved the restart logic to the Main() method, this ensures that it is always reached. I have also added a Boolean “IsRestart” that is used to indicate a restart by the “Restart()” method.

### DIFF
--- a/Server/MainServer.cs
+++ b/Server/MainServer.cs
@@ -38,6 +38,8 @@ namespace Server
 
         public static readonly CancellationTokenSource CancellationTokenSrc = new CancellationTokenSource();
 
+        private static bool IsRestart = false;
+
         public static void Main()
         {
             try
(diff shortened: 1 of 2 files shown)


DasSkelett (approved):
Very nice, thanks alot!
The server restart logic isn't great, it doesn't work on any OS but Windows. It will need more rework eventually, but at least now it works reliably on Windows again.

PR #544, opened 2023-02-13
Add dedicated server

mlltAero:
Add Ghost Server to dedicated server list.

### DIFF
--- a/MasterServersList/DedicatedServersList.txt
+++ b/MasterServersList/DedicatedServersList.txt
@@ -39,3 +39,6 @@ zandstra.duckdns.org:8800
 ksp.delabonifier.com:8800
 ksp.delabonifier.com:8820
 ksp.delabonifier.com:8830
+#Ghost Server
+ksp.descaltr.com:8800
+78.58.229.204:8800


DasSkelett:
`ksp.descaltr.com` doesn't seem to be set up correctly, it doesn't return any A records.

DasSkelett:
Obligatory question for dedicated servers: have you read through https://github.com/LunaMultiplayer/LunaMultiplayer/wiki/Dedicated-server and are you planning to keep it running 24/7 for a long time?

mlltAero:
Right. Just realized that my A record is set up incorrectly.

mlltAero:
> Obligatory question for dedicated servers: have you read through https://github.com/LunaMultiplayer/LunaMultiplayer/wiki/Dedicated-server and are you planning to keep it running 24/7 for a long time?

Yes, I am planning to run it 24/7.

DasSkelett:
Sorry for the delay. The domain is fixed now indeed.
The domain however points to the same IP address as the one you added in the second line. The same address twice is unnecessary, can you remove the second one?

PR #551, opened 2023-06-01
Port changes for DeltaServers in DedicatedServersList.txt

JanePixel:

Due to an upcoming network infrastructure change, the ports on which these servers run needs to change.
I have not yet changed the ports on the servers, so they stay marked as dedicated until this PR is merged.


DasSkelett (commented):
Thanks for the update!

PR #554, opened 2023-06-24
Feature: Prometheus Metrics Framework

celestialorb:
This is a revamp of an old PR (#479) of mine. I'm getting back into KSP and would like a server that can export metrics in a way that Prometheus can scrape and capture them, allowing me to easily view trends over time in something like Grafana.

I decided to create a new branch and a new PR to avoid cleaning up the old one and to refresh my own memory as to how this works.

This initial PR just establishes a basic endpoint for Prometheus metrics and only exposes two metrics: one for total online player count as well as one for online status of individual players (if detailed player metrics are enabled).

If this is accepted I will create PRs to add individual groups of metrics to the server.
Looking at what I've done in the past for an old server of mine this would/could include:
- Space Program Metrics
  - Funds
  - Science
  - Reputation
- Vessel Metrics
  - Position
  - Orbital Elements
  - Staging
  - Resource Levels
- Facility Metrics
  - Levels
  - maybe damage?
- Contract Metrics
  - State of available contracts (i.e. rewards, accepted/offered)
- Technology Metrics
  - Science cost
  - Number of parts
- Subspace Metrics
  - Time difference of each subspace from server
- Kerbal Metrics
  - Attributes (bravery, dumb, etc.)
  - Flight Counts
  - Deaths
  - Recoveries

### DIFF
--- a/Server/Server/Metrics/Player.cs
+++ b/Server/Server/Metrics/Player.cs
@@ -8,10 +8,10 @@ namespace Server.Metrics {
             new Prometheus.GaugeConfiguration{}
         );
 
-        public static readonly Prometheus.Counter Online = Prometheus.Metrics.CreateCounter(
+        public static readonly Prometheus.Gauge Online = Prometheus.Metrics.CreateGauge(
             "lmp_player_online",
             "Whether or not the player is currently online.",
-            new Prometheus.CounterConfiguration{LabelNames = new[] {"name"}}
+            new Prometheus.GaugeConfiguration{LabelNames = new[] {"name"}}
         );
 
         public static void AddPlayer(string name) {
--- a/Server/Server/Metrics/Player.cs
+++ b/Server/Server/Metrics/Player.cs
@@ -17,7 +17,7 @@ namespace Server.Metrics {
         public static void AddPlayer(string name) {
             Total.Inc(1);
 
-            if (MetricsSettings.SettingsStore.DetailedPlayerMetrics) {
+            if (MetricsSettings.SettingsStore.EnablePlayerDetailedMetrics) {
                 Online.WithLabels(name).IncTo(1);
             }
         }
@@ -25,7 +25,7 @@ namespace Server.Metrics {
         public static void RemovePlayer(string name) {
             Total.Inc(-1);
 
-            if (MetricsSettings.SettingsStore.DetailedPlayerMetrics) {
+            if (MetricsSettings.SettingsStore.EnablePlayerDetailedMetrics) {
                 Online.RemoveLabelled(name);
             }
         }
--- a/Server/Server/Metrics/Player.cs
+++ b/Server/Server/Metrics/Player.cs
@@ -1,0 +1,13 @@
+namespace Server.Metrics {
+    public class Player {
+        public static readonly Prometheus.Counter Count = Prometheus.Metrics.CreateCounter(
+            "lmp_player_online",
+            "Whether or not the player is currently online.",
+            new Prometheus.CounterConfiguration{LabelNames = new[] {"name"}}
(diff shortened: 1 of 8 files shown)


celestialorb:
Tagging @gavazquez and @DasSkelett as they were involved in the review of the original PR.

celestialorb:
Looks like [the Appveyor jobs are failing](https://ci.appveyor.com/project/gavazquez/lunamultiplayer/builds/47386204/job/ki0t60x0f775vym2). I guess the secret needs to be updated?

Server/Server/Metrics/Player.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Server.Metrics {
+    public class Player {
+        public static readonly Prometheus.Gauge Total = Prometheus.Metrics.CreateGauge(
+            "lmp_player_online_total",
+            "The total count of unique players online.",
+            new Prometheus.GaugeConfiguration{}
+        );
+
+        public static readonly Prometheus.Counter Online = Prometheus.Metrics.CreateCounter(
```
DasSkelett, line 11: As far as I remember the Prometheus recommendations, this should probably also be a gauge as it isn't increasing continuously.

Server/Web/WebServer.cs
```diff
@@ -65,6 +65,18 @@ public static void StartWebServer()
...
+                        // If the Prometheus endpoint is enabled, build and add the HttpRouter to handle it to the server.
+                        if (MetricsSettings.SettingsStore.Enabled) {
+                            LunaLog.Info("Enabling Prometheus endpoint");
+                            Server.Use(
+                                new HttpRouter().With(
+                                    MetricsSettings.SettingsStore.Endpoint,
+                                    new MetricsHandler()
+                                )
+                            );
+                        }
```
DasSkelett, line 78: Instead of creating a separate `HttpRouter` for this path, I think it would be a bit cleaner to add the metrics sub-handler to the existing one (which does require a bit of restructuring due to the if-clause), something like this?

(`HttpRouter.With()` does modify itself, so we are fine ignoring the returned object).

```suggestion
                        var httpRouter = new HttpRouter();
                        Server.Use(httpRouter);
                        httpRouter.With(string.Empty, new RestHandler<ServerInformation>(new ServerInformationRestController(), JsonResponseProvider.Default));

                        // If the Prometheus endpoint is enabled, add it to the HttpRouter.
                        if (MetricsSettings.SettingsStore.Enabled) {
                            LunaLog.Info("Enabling Prometheus endpoint");
                            httpRouter.With(
                                MetricsSettings.SettingsStore.Endpoint,
                                new MetricsHandler()
                            );
                        }
```

DasSkelett (commented):
Sorry for the long wait. Thanks for the work, looks very good, two small comments:

Server/Server/Metrics/Player.cs
celestialorb: Yeah, that's correct. This should definitely be a gauge, not sure why I created it as a counter.

Server/Web/WebServer.cs
celestialorb: Yeah, it would be cleaner if we could avoid the creation of another HttpRouter object.

celestialorb:
Added the changes recommended @DasSkelett!

celestialorb:
@DasSkelett Shall this be merged in, or would you like to fix the AppVeyor build first?

celestialorb:
@DasSkelett Just a friendly poke on this PR.

BraveCaperCat2 (approved):
It all looks good to me, besides the merge conflicts from the PR being over 3 years old.

PR #559, opened 2023-08-29

zhouyiqing0304:


DrewBanyai:
Dedicated server entries require both an IP and a Port. Closing PR.

PR #560, opened 2023-08-29

zhouyiqing0304:
ok i know. the server will works for 10years.it doesn't have any password


DrewBanyai:
Dedicated server entries require both an IP and a port. Closing PR.

PR #568, opened 2023-12-17
Add kerbal.io xmas server

emlynlj:
Heya Skelett! Another year passes... another christmas event comes around...

I've just replaced the old one with the new hostname (server.kerbal.io).

This time I'll remember to delete it after the event's over :).
It will remain password protected for just a day or two until it launches on the 20th, hope that's ok.


DrewBanyai:
Neither server was accessible at time of testing. Original will be removed on another branch. New one will not be added. Closing PR.

PR #577, opened 2024-09-04
Upgrade to .NET 8 LTS, Docker images to alpine:3.20

DasSkelett:

- Upgrade server & masterserver to .NET 8 LTS
- Upgrade Docker images to alpine:3.20


DrewBanyai:
It would have been difficult to confirm 100% that this implementation of a Dotnet 8 upgrade was correct and we would want to update to Dotnet 9 or higher at this point anyway, so I went ahead and did that and built another PR: https://github.com/LunaMultiplayer/LunaMultiplayer/pull/615

I've been testing it for about a month now, I'm not sure I can say it's actually faster but builds are a lot simpler to do (don't need a direct call to msbuild or a full install of Visual Studio, for one...) and it's more in line with modern standards.

If it goes through, I'll be deleting this PR. I'm sorry it seems to have fallen by the wayside. You did some good work, but bandwidth to look through each Issue and PR in this was not available until recently. I'm looking to modernize the entire project a bit so I'm going through what I can this month.

PR #592, opened 2025-08-23
Fix kerbals not being properly matched by their name

### LINKED ISSUE #541: Kerbals whose (first) name is a substring of another Kerbal disappear
Admittedly the title of the issue is a guess, but I'd also call it an educated guess.  Took us 3 hours to figure out *something* that would un-bork our save and deleting Billy-Bobdas Kerman was what did it.

I looked and it seems I don't have a log from when this happened.  I can reproduce the issue if necessary, but I hope the rest of the report is sufficient information (we debugged the issue as well as we could and found the cause and worked around it...and then kept playing and lost the logs from the now-deleted broken instances).

---

Version:

Latest release (0.29.0) for client and server

---

> If you have access to the server. Please zip the whole "Universe" and "Config" folder.
Also if you are running mods in the server, attach the "LMPModControl.txt"

I don't currently have the server in that state but if necessary I can return it to that state and upload the requested files.

Here is the [config folder](https://github.com/LunaMultiplayer/LunaMultiplayer/files/10497076/Config.zip) (which is unchanged from before the issue) in case it's useful.

---

Background:

- OS: Linux (server), Linux and Windows (clients)
- KSP Version: build 10132464 (Steam latest)
- LMP version (include the build number if you are using a nighly build): see above

---

Mods (remember that **mods are not supported**):  

RSS and most of its suggested mods from CKAN (not RO, also RSS with stock parts is hard)

Complete list:

```
ls -1 KSPFolder/GameData

000_Harmony
000_KSPBurst
Custom_FARAeroData.cfg
EnvironmentalVisualEnhancements
Kopernicus
KSCSwitcher
KSPCommunityFixes
LMP Readme.txt
LunaMultiplayer
ModularFlightIntegrator
ModuleManager.4.2.2.dll
ModuleManager.ConfigCache
ModuleManager.ConfigSHA
ModuleManager.Physics
ModuleManager.TechTree
Parallax
Parallax_StockTextures
RealSolarSystem
RSSDateTime
RSS-Textures
RSSVE
Scatterer
ScattererAtmosphereCache
SpaceTuxLibrary
Squad
SquadExpansion
```

---

Expected behaviour:  

Kerbals save and load correctly.

---

Current behaviour:  

If a contract (or the astronaut complex) creates a Kerbal, it assigns it a name.  Some names seem to be problematic for an unknown reason.  We accepted a tourist contract which included a tourist named Billy-Bobdas Kerman, but then called it a night.  When next we loaded the save, Bill and Bob Kerman were nowhere to be found (not MIA or anything) and the vessel Bob was on caused an infinite loading screen if you attempted to fly it.  Bill's and Bob's `Kerbals/name.txt` files were also gone.  We discovered that if we restored from that good save and deleted the tourist contracts and Billy-Bob Kerman's file that this issue went away.  We're *pretty sure* this is the issue and have not seen it since we weighted low and now decline all Kerbal-generating contracts.

---

Steps to reproduce:  

See "current behavior" section (I couldn't find a clean separation between this and that section in this case).

This: [persistent.txt](https://github.com/LunaMultiplayer/LunaMultiplayer/files/10496976/persistent.txt) is the (renamed because github doesn't recognize .sfs) save file which we were able to restore from.  This file is pre-"fix".  Load it onto a server, join, fly the earth orbit science lab if you want (I did this when testing and it worked on first load), leave the server, restart the server, rejoin and now Bill and Bob are gone and that science lab will hit you with an infinite load screen.

---

I've poked around in the source a bit in hopes to find something which stood out at me that could cause this issue, but not for too long yet.  I'll probably keep looking though, and will post back here if I find something, and open a PR if I find something definitive that I can reasonably fix.

Matszwe02:
Modify KerbalSystem to match kerbals by their exact name match

Add `existingKerbal` and `existingProtoCrew` for direct kerbal access instead of operating on list

Directly match kerbal's name with `k.name == protoCrew.name` as current approach did allow non exact matches

fixes #541

### DIFF
--- a/LmpClient/Systems/KerbalSys/KerbalSystem.cs
+++ b/LmpClient/Systems/KerbalSys/KerbalSystem.cs
@@ -3,6 +3,7 @@ using KSP.UI.Screens;
 using LmpClient.Base;
 using LmpClient.Events;
 using System.Collections.Concurrent;
+using System.Linq;
 using System.Reflection;
 using UnityEngine;
 
@@ -136,7 +137,11 @@ namespace LmpClient.Systems.KerbalSys
                 var refreshDialog = false;
                 while (KerbalsToRemove.TryDequeue(out var kerbalNameToRemove))
                 {
-                    HighLogic.CurrentGame.CrewRoster.Remove(kerbalNameToRemove);
+                    var kerbalToRemove = HighLogic.CurrentGame.CrewRoster.Crew.FirstOrDefault(k => k.name == kerbalNameToRemove);
+                    if (kerbalToRemove != null)
+                    {
+                        HighLogic.CurrentGame.CrewRoster.Remove(kerbalToRemove);
+                    }
                     refreshDialog = true;
                 }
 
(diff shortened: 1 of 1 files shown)


Matszwe02:
I have an update on this: somehow my tests didn't catch that the issue still appears - I've narrowed it down to `Bob` and `Bill` kerbals being removed from server files the same moment a new kerbal with overlapping names is generated by the server.
My tests didn't fully catch that, because I was inserting these bad kerbal names into Universe, which isn't generating any issues. That makes further debugging much harder, as it only breaks the moment server chooses to create a bad kerbal.

So I'm still investigating the bug.

Issue #587, opened 2025-05-08
Cannot add settings in gameplaysettings.xml


OPenworldgamer:
Background:

OS: Windows
KSP Version: 1.12.5
LMP version (include the build number if you are using a nighly build): 0.29.0
Mods (remember that mods are not supported):

For example:
Click-Through-Blocker
and a lot more

Expected behaviour:
Modifying the settings xml with new settings applies those settings.

Current behaviour:
Settings in the xml file that I try to custom add get deleted. The goal is that I could change mod settings and other game settings so they don't keep getting reset.

Steps to reproduce:
-Change gameplaysettings.xml
-start server-the new settings are erased

DrewBanyai:
If you're still interested in this, please let me know what settings you are trying to add and what each should do. It's likely you would have to add custom code to your server to pull in these values and apply them to how things work on the server. It sounds as if you're just hoping it picks up what it's supposed to do based on whatever you're naming these variables...

BraveCaperCat2:
> If you're still interested in this, please let me know what settings you are trying to add and what each should do. It's likely you would have to add custom code to your server to pull in these values and apply them to how things work on the server. It sounds as if you're just hoping it picks up what it's supposed to do based on whatever you're naming these variables...

I believe this person is experiencing a problem where the CTB settings keep getting reset for them every time they join an LMP server. It happens to me too.
They just tried to fix it on their server by adding the CTB setting to the LMP server's config. (which of course did not work)

PR #594, opened 2025-08-26
Fix game parameter syncing to not overwrite unrelated existing game parameters

iddqd0:
Fixes the constant resetting of parameters set by mods like CTB (Click-Through-Blocker) every time you connect to a server.
Probably solves the issue encountered in #587.

- Overwrite only predefined sections of parameters instead of the entire block.


iddqd0:
Upon further testing - this still doesn't actually seem to be working. Changing to draft for now.

PR #619, opened 2026-04-13
Various fixes for v0.29.1

DrewBanyai:
## Overview

This PR collects bug fixes, thread-safety improvements, and version/build infrastructure changes targeting the v0.29.1 release. The version was initially bumped to 0.30.0 and then walked back to 0.29.1 to preserve messaging compatibility with existing clients while the new `IsCompatibleWithPeer` system matures.

---

## Version & Build

- Version markers updated across all `.cs` and `.yaml` files (0.29.0 → 0.29.1; started at 0.30.0, reverted to 0.29.1 to avoid incompatibility until the peer-compatibility system is adopted).
- Added a single build script that compiles LmpClient, Server, and MasterServer in both debug and release configurations in one pass.
- Updated Harmony dependency to **2.2.1**; made external dependency file discovery recursive so the new Harmony DLL paths are picked up correctly.
- Merged upstream CI improvements: appveyor `dotnet publish` updates, devcontainer config, and `Dockerfile_Server` changes.

## Networking & Compatibility

- Added `LmpVersion` to `ClientStructure` and surfaced it through the Handshake system, making the client version visible to the server.
- Introduced `IsCompatibleWithPeer`: a forward-looking compatibility check that will allow specified version groups (e.g. ≥ 0.29.1, including a future 0.30.0) to co-exist on the same server without strict version equality, once the playerbase has migrated.
- Replaced the deprecated `Common.PlatformIsWindows()` call in MainServer with `OperatingSystem.IsWindows()`.
- Fixed a race condition in the Handshake system where a shared mutable reason string could be corrupted when multiple clients connected concurrently; replaced with a per-call output string parameter.
- Fixed a race condition in `ChatMessageHandler` where multiple chat messages arriving in rapid succession could corrupt the shared `StringBuilder` state, causing a disconnection or exception; the handler now resets the builder safely before each use.

## Vessel & Physics

- Position sync event now fires immediately when a player takes control of a vessel, and simultaneously writes the correct body name into the vessel's ORBIT data.
- `Orbit.IDENT` is now assigned during controlled flight, ensuring it is correctly persisted to the vessel text file
- Fixed `InternalGetMessageSize` in `VesselUpdateMsgData`: the size calculation was missing the three `Com` float values, causing message sizing mismatches.
- Vessels added to the destruction queue are now immediately deactivated, preventing Unity components from continuing to execute on a vessel that is pending removal.

## Thread Safety & Performance

- `RemovedVessels` collection converted to a thread-safe structure, eliminating lock contention and reducing lookup complexity from O(n) to O(1).

## Contracts & Miscellaneous

- Various fixes for Contract loading and management: handles edge cases in contract state restoration and player lock acquisition.
- Added support for unknown/missing mod parts so that vessels with unrecognised part modules no longer crash the contract or vessel system.


gavazquez:
Amazing PR! Really everything there was needed :)

gavazquez:
Considering that you're really good at messaging and so on you may have a look at the vesseldock system. It was a total mess to know what players should be in control of what vessel and the delays and so on caused a lot of problems. Such as 2 players controlling the same ship and so on

DasSkelett:
Thanks! For the next ones, can you please make sure to properly rebase again, so there are no "Merge master into ..." commits? They tend to mess up the history a bit.

DrewBanyai:
Oof. I'll try to remember next time.

PR #625, opened 2026-04-16
Fix: NaN orbit no longer permanently deletes vessel from server

AdmiralRadish:
When a vessel's orbit calculation produces NaN (typically from patched-conics numerical instability during SOI transitions), the current code sends VesselRemove to the server and calls KillVessel with addToKillList=true. This permanently deletes the .vcproto file and adds the vessel to the in-memory kill list, making it irrecoverable for all players without a server restart and manual file restoration.

The NaN is a client-side calculation artifact -- the server's stored data is almost always valid. This change removes the SendVesselRemove call and passes false to KillVessel, so the vessel is cleaned up locally but the server retains it. Other clients are unaffected, and the vessel reloads normally on reconnect.

Normal vessel destruction (crashes, overheating, overpressure, G-forces) is unaffected -- those go through Part.Die / Vessel.Die / onVesselWillDestroy, a completely separate code path.


AdmiralRadish:
@DrewBanyai I'm going to open a number of items based on the fixes we are doing for our live server. Any of them are open for discussion, feel free to reach out on any item at all!

PR #635, opened 2026-04-16
Sanitize persistent IDs before vessel load to prevent freezes

AdmiralRadish:
KSP assigns every vessel and part a `persistentId` that must be unique within the running game. When LMP loads a vessel from the server, the incoming proto often carries IDs that already exist locally (because the vessel was loaded before, or another client chose the same IDs). KSP handles this with `HandlePartPersistentIdCollision`, but that handler fires **per conflicting part on the main thread** — and under concurrent LMP vessel loads (multiple vessels arriving in the same frame), the collision cascade causes multi-second freezes proportional to vessel part count.

This adds a `SanitizePersistentIds` pass before `ProtoVessel.Load()` that checks every vessel-level and part-level `persistentId` against the live `FlightGlobals` registries (`PersistentVesselIds`, `PersistentLoadedPartIds`, `PersistentUnloadedPartIds`) and remaps any collisions using `FlightGlobals.GetUniquepersistentId()`. The collision handler never fires because KSP sees clean IDs.

The incoming IDs are transient transport values that only need to be unique on this client — the server's save is authoritative, so remapping is safe. Logs each remap for observability.

Upstream recently added zombie-vessel cleanup and a try/catch around `Load()` in the same file — those changes are in the full-reload path while this sits immediately before the `Load()` call and in a new region at the bottom. No overlap.

### DIFF
--- a/LmpClient/VesselUtilities/VesselLoader.cs
+++ b/LmpClient/VesselUtilities/VesselLoader.cs
@@ -66,12 +66,7 @@ namespace LmpClient.VesselUtilities
                 LunaLog.Log($"[LMP]: Loading vessel {vesselProto.vesselID}");
             }
 
-            // Vessel.Start() calls Vessel.RebuildCrewList() which iterates protoModuleCrew on every
-            // ProtoPartSnapshot.  A null slot in that list causes a NullReferenceException that Unity
-            // catches internally (logged as [EXC]) — it never reaches our catch block below.  Strip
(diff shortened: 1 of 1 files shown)


AdmiralRadish:
absorbed the upstream crew null-strip into [SanitizePersistentIds] method (it belongs there), kept upstream module null-check block untouched, rebased and force-pushed

PR #638, opened 2026-04-16
Fix scenario ConfigNode parsing and add error handling

AdmiralRadish:
KSP's ConfigNode.WriteNode() wraps unnamed nodes in outer { } braces when serializing. LunaConfigNode's parser treats braced content as an unnamed child node, so when the server tries to do GetValue() on the parsed result it gets null back — and the scenario update silently does nothing. This has been quietly breaking contract, science, technology, strategy, and achievement sync on the server for any client that sends data through the normal share-progress path.

This adds a shared ParseClientConfigNode() helper in ScenarioBaseDataUpdater that strips the outer braces before parsing, and uses it in all six scenario data updaters. The same stripping is applied in LoadExistingScenarios (for scenario files read from disk) and GetScenarioInConfigNodeFormat so the format is consistent everywhere.

Every GetValue() call that previously assumed non-null (and would NRE on malformed data) now null-checks first. Every Task.Run body is wrapped in try/catch with LunaLog.Error so exceptions don't silently swallow scenario updates anymore.

Also fixes a logic bug in the strategy updater where AddNode was being called unconditionally after the FirstOrDefault search instead of only when no existing node matched — so duplicate strategies would pile up in the scenario file.

The contracts updater incorporates the recent upstream finished-contract handling (moving completed/failed/cancelled contracts to CONTRACTS_FINISHED) with the brace-stripping fix applied on top.

### DIFF
--- a/Server/System/Scenario/ScenarioAchievementsDataUpdater.cs
+++ b/Server/System/Scenario/ScenarioAchievementsDataUpdater.cs
@@ -1,6 +1,6 @@
 using LmpCommon.Message.Data.ShareProgress;
-using LunaConfigNode.CfgNode;
-using System.Text;
+using Server.Log;
+using System;
 using System.Threading.Tasks;
 
 namespace Server.System.Scenario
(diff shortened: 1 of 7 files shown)


JurassikLizard:
This would be a very useful PR, as currently it seems to me that the contracts on my server are stored as completed, but when syncing to the client it does not work.
Note: I believe this is useful for instances with modded contracts, which sometimes are formatted differently than stock ones.

PR #639, opened 2026-04-16
Fix: Validate docking port FSM state before processing undocking

AdmiralRadish:
When LMP processes a remote undock message it finds the docking port part and calls Undock() on it directly. The problem is that docking port FSMs regularly end up in the wrong state in multiplayer — proto reloads from the network reset the FSM without going through KSP's normal docking sequence, so a port that should be in docked state might actually be sitting in a different state. Calling Undock() on a port in that state either silently fails or throws.

This adds a check in VesselUndock.ProcessUndock() before the undock call. If the port's FSM is properly docked, proceed normally. If it's stuck in a recoverable transient state (Disengage, Acquire, etc.), attempt to recover the FSM first using the same fsm.StartFSM() technique that DockRotate uses, then proceed. If it's in Ready or some other non-docked state, skip the undock entirely with a log warning since the port clearly isn't docked.

The heavy lifting is in a new DockingPortUtil utility class that handles state classification, walks the part tree's attach nodes to find the other side of a docking pair with fallback to dockedPartUId lookup, docker/dockee role inference, and pair recovery.

There's also a FixDockingPortFsmStates() method that does a comprehensive vessel-wide scan for broken ports: ports where the serialized state says docked but the FSM disagrees, ports with a partner UID but wrong FSM state, ports that are physically connected in the part tree but the FSM doesn't reflect it, ports stuck in transient states with no partner, and ghost docked states where the FSM says docked but no physical partner exists on the vessel. It also cleans up stale cross-vessel dockedPartUId references left over from previous undocks.

### DIFF
--- a/LmpClient/Harmony/Vessel_GoOffRails.cs
+++ b/LmpClient/Harmony/Vessel_GoOffRails.cs
@@ -1,21 +1,41 @@
-﻿using HarmonyLib;
+﻿using System.Linq;
+using HarmonyLib;
 using LmpClient.Events;
+using LmpClient.VesselUtilities;
 
 // ReSharper disable All
 
 namespace LmpClient.Harmony
 {
     /// <summary>
-    /// This harmony patch is intended to trigger an event when FINISHED unpacking a vessel
+    /// This harmony patch is intended to trigger an event when FINISHED unpacking a vessel.
+    /// Also repairs docking port FSM states that fail to restore correctly after vessel loading.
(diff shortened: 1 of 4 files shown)


AdmiralRadish:
There may still be some edge cases I am missing if a docking port is broken, then an engineer moves it across the vessel, and the broken pair port docks again, etc. I'm trying not to get caught up in those extreme niche cases quite yet.

PR #642, opened 2026-04-16
Fix: Suppress first-run popups in multiplayer

AdmiralRadish:
LMP rebuilds the game object from scratch on every server connection and excludes ScenarioNewGameIntro from scenario sync. This means the tutorial flags (kscComplete, editorComplete, tsComplete) are always false when loading, so the KSC welcome popup, VAB tutorial, and tracking station tutorial re-trigger every session.

ScenarioNewGameIntro_OnLoad.cs is a Harmony prefix on ScenarioNewGameIntro.OnLoad that sets all three tutorial flags to true and skips the stock OnLoad body. This prevents the KSC welcome popup from spawning and marks all tutorials as complete for the session.

HarmonyPatcher.cs adds a PatchOptionalMods() call after PatchAll, with an initial entry for ClickThroughBlocker. CTB's OneTimePopup.Start gets re-triggered on every server join despite PopUpShown.cfg already being set, presumably because LMP's fresh Game creation confuses CTB's first-run detection. The patch looks up the type by name at runtime so it's safely a no-op if CTB isn't installed.

IgnoredScenarios.cs adds ScenarioNewGameIntro to IgnoreReceive. It was already in IgnoreSend with a matching comment, but not in IgnoreReceive — meaning a scenario packet from the server could still overwrite the local flags back to false.

### DIFF
--- a/LmpClient/Base/HarmonyPatcher.cs
+++ b/LmpClient/Base/HarmonyPatcher.cs
@@ -1,5 +1,5 @@
-﻿using System.Reflection;
-
+﻿using System;
+using System.Reflection;
 namespace LmpClient.Base
 {
     public static class HarmonyPatcher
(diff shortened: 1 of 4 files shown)


DrewBanyai:
Closing this one. We don't have the problem described here except for CTB and I'm going to implement that change manually instead of bringing this whole PR into master.

PR #644, opened 2026-04-16
Adding more safeguards to stop null reference exceptions in loading or updating vessels

DrewBanyai:


DrewBanyai:
## Summary

Network-loaded vessels can arrive with incomplete data (missing mods, partially-reconstructed crew lists, null part modules) that causes KSP to throw `NullReferenceException` deep inside its own internals — places our `try/catch` blocks never reach. The result is hundreds of `[EXC]` log entries per second, visible framerate drops, and in some cases the UI freezing on save.

This PR adds several layered safeguards to catch and recover from these situations cleanly.

---

## Changes

### `VesselLoader.cs` — Pre-load and post-load validation

**Pre-load:** Before calling `ProtoVessel.Load()`, any `null` entries in each `ProtoPartSnapshot.protoModuleCrew` list are stripped. `Vessel.RebuildCrewList()` (called from `Vessel.Start()`) iterates this list and throws internally on nulls — Unity catches it as `[EXC]`, so it never reaches our code.

**Post-load:** After the vessel loads successfully, every part and module slot is checked for nulls. If a bad slot is found (e.g. from a mod the client doesn't have), the vessel is immediately torn down and `false` is returned. This prevents a broken vessel from ever reaching `Vessel.FixedUpdate()` where `UpdateCaches()` would throw on every physics tick.

### `Harmony/Vessel_UpdateCaches.cs` — Physics-tick safety net (new file)

A two-part Harmony patch on `Vessel.UpdateCaches()`:

- **Prefix:** If the vessel's ID is already in the kill set, the method is skipped entirely until Unity destroys the `GameObject`. This prevents any further exceptions from a vessel that's already been condemned.
- **Finalizer:** If `UpdateCaches()` throws a `NullReferenceException`, the vessel is queued for removal via `VesselRemoveSystem` and the exception is suppressed (returning `null` to Harmony stops Unity from logging `[EXC]`). Only the first occurrence is acted on, so log output stays clean.

### `Harmony/VesselModule_Start.cs` — VesselModule startup safety net (new file)

A Harmony finalizer on `VesselModule.Start()`. Known problem cases include `CometVessel.OnStart()` (null comet discovery/orbit data) and `SuspensionLoadBalancer.OnStart()` (null wheel-module reference) on vessels loaded over the network. The finalizer intercepts `NullReferenceException`, logs a single diagnostic warning, and suppresses the exception so the rest of vessel initialization can continue.

---

## Why Harmony finalizers instead of pre-validation?

Pre-validating every field that `UpdateCaches()` or `OnStart()` might dereference is inherently incomplete — the call graphs touch parts, modules, `VesselModules`, resources, transforms, and more. Harmony finalizers catch every case uniformly, regardless of which internal field is null, and they run even when the original method threw before returning.

Issue #380, opened 2020-12-08
Docking with other player leads to explosions.


jeq1:
Remember to attach the "KSP.log" and the "output_log.txt" whenever is possible!  
You can find the KSP.log in ".\Kerbal Space Program\KSP.log"  
And output_log.txt in "C:\Users\YOURUSERNAME\AppData\LocalLow\Squad\Kerbal Space Program\output_log.txt"
(you probably need to display hidden folders to get into AppData)

In case you are using a nighly build of LMP from appveyor include the build number.
For example: 0.0.9.**384**
Otherwise, specify the version of LMP you are using from the github releases.

**If you have access to the server. Please zip the whole "Universe" and "Config" folder.
Also if you are running mods in the server, attach the "LMPModControl.txt"**

------------------------------------------------------------------------------
------------------------------------------------------------------------------
------------------------------------------------------------------------------

OS: 

------------------------------------------------------------------------------

KSPVersion: 
1.10.1
------------------------------------------------------------------------------

LMP version (include the build number if you are using a nighly build): 
0.26.0
------------------------------------------------------------------------------

MODS. Remember that **mods are not supported**: 

------------------------------------------------------------------------------

Expected behaviour: 
dock while friend controls other vessel and voila you have space station
------------------------------------------------------------------------------

Current behaviour: 
Docking ports explodes and tourists are gone.
------------------------------------------------------------------------------

Steps to reproduce:

lfuelling:
I tried reproducing this with LMP 0.27 and KSP 1.11, docking seems to work without explosions although there is weird wobbling (more than usual) when the docking happens with momentum not aligned with the docking axis.

jeq1:
Now it doesn't explode, but game crashes for one of docking players to desktop.

ShiralynDev:
[KSP.log](https://github.com/user-attachments/files/27590910/KSP.log)

Tested this in latest master, leads to a kick, no explosions tho. I will rename the issue since the explosion seems to be fixed. I think an explosion could happen due to the magnets in the docking ports doing weird stuff when both clients are trying to sync it.

ShiralynDev:
Marked as enhancement since with a kick it works but would be to start spectating when docking with another player.

BraveCaperCat2:
> Marked as enhancement since with a kick it works but would be to start spectating when docking with another player.

How is this an enhancement?
[...]

PR #660, opened 2026-04-24
fix: vessel coupling system for synchronized docking

Fierce-Cat:

- **Fixed `NullReferenceException` crashes during failed docking:** Prevents the game from hard-crashing or breaking the client's camera if a coupling/docking event fails to execute due to glitchy parts or missing vessels.
- **Fixed loss of vessel control after docking:** Ensures that the player who actively initiates a docking maneuver (e.g., docking to a space station) immediately gains all controls of the resulting combined vessel, preventing them from being locked out of their own ship.
- **Fixed visual desyncs and physics bugs for spectators:** Resolves issues where other players observing a docking event would see parts floating in space or ships exploding, by forcing their clients to completely rebuild the newly docked vessel.

- **`VesselCouple.cs`**: Wrapped `AfterCouplingEvent()` in a `if (coupleResult)` check to ensure it only fires on successful docking. Added explicit null cleanup for `_dominantVessel` and `_weakVessel` when docking fails to prevent stale data corruption, and added a null-check safeguard before calling `.MakeActive()`.
- **`VesselCoupleEvents.cs`**: Removed the conditional check for prior lock ownership. The code now forcefully acquires the `ControlLock`, `UpdateLock`, and `UnloadedUpdateLock` on the combined vessel for the player triggering the docking event.
- **`VesselProtoEvents.cs`**: Updated `System.MessageSender.SendVesselMessage(partFrom.vessel)` to include `forceReload: true` inside `PartCoupled()`, explicitly commanding other connected clients to rebuild the vessel model from scratch rather than attempting a standard interpolation update.

### DIFF
--- a/LmpClient/Systems/VesselCoupleSys/VesselCouple.cs
+++ b/LmpClient/Systems/VesselCoupleSys/VesselCouple.cs
@@ -1,4 +1,4 @@
-﻿using HarmonyLib;
+using HarmonyLib;
 using LmpClient.Extensions;
 using LmpClient.Systems.VesselRemoveSys;
 using LmpClient.VesselUtilities;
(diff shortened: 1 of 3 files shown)


ShiralynDev:
Believe this would solve #380.

PR #661, opened 2026-04-26
Feature: Server Version Filter

BraveCaperCat2:
Add a basic "version filter" to the master server. This will prevent people from registering servers on the server list if they're not using one of the actual LMP versions.
Sending a message back to the server owner to inform them of this is a work in progress.
Additionally, everything in #658 because those commits were included in the branch I've been working on this in

### DIFF
--- a/LmpMasterServer/Lidgren/MasterServer.cs
+++ b/LmpMasterServer/Lidgren/MasterServer.cs
@@ -34,6 +34,12 @@ namespace LmpMasterServer.Lidgren
             new TimeoutConcurrentDictionary<IPAddress, object>(1000);
         private static MasterServerMessageFactory MasterServerMessageFactory { get; } = new MasterServerMessageFactory();
 
+        /// <summary>
+        /// Checks if the server version provided is less than or equal to the version of the master server.
+        /// Later versions will be invalid (false return value) unless specified as compattible in CrossCompatibleVersionLines.
+        /// </summary>
+        public static bool IsValidServerVersion(Version ServerVersion) => (ServerVersion <= LmpVersioning.CurrentVersion) || LmpVersioning.IsCompatible(ServerVersion);
+
         public static async Task StartAsync()
         {
             var config = new NetPeerConfiguration("masterserver")
@@ -254,6 +260,11 @@ namespace LmpMasterServer.Lidgren
         private static void RegisterServer(IMessageBase message, NetIncomingMessage netMsg)
         {
             var msgData = (MsRegisterServerMsgData)message.Data;
+            if (!Version.TryParse(msgData.ServerVersion, out var serverVersion) || !IsValidServerVersion(serverVersion))
+            {
+                LunaLog.Debug($"Server registration rejected for server at {netMsg.SenderEndPoint} with provided version \"{msgData.ServerVersion}\". (not a valid server version)");
+                return;
+            }
 
             if (ServerDictionary.TryGetValue(msgData.Id, out var existing))
             {


LmpMasterServer/Lidgren/MasterServer.cs
```diff
@@ -34,6 +34,57 @@ public class MasterServer
...
+            "0.26.0",
+            "0.27.0",
+            "0.28.0",
+            "0.29.0",
+            // Current version
+            "0.29.1",
+            // Future versions (>0.29.1)
+            "0.29.2",
+            "0.30.0"
+        };
```
DasSkelett: I'm not too fond of having another place we need to change/add the version number on new releases.
Could we instead just to a plain "<= current release" check using a properly parsed version object and `LmpVersioning`?

BraveCaperCat2: @DasSkelett
I guess that'd work

BraveCaperCat2: Resolved in the latest commit

BraveCaperCat2:
@DasSkelett the build succeeded this time
The only issue I can spot is that `msgData.ServerVersion` could (theoretically) error when parsing (see line 264) but it's set server-side by `LmpVersioning.CurrentVersion.ToString(3)` so it should always parse without errors, as long as the data doesn't get corrupted in transit.

DasSkelett:
Sorry for the long wait, busy weeks...
Yeah I think this is a good approach, thanks for doing the change.

> The only issue I can spot is that `msgData.ServerVersion` could (theoretically) error when parsing (see line 264) but it's set server-side by `LmpVersioning.CurrentVersion.ToString(3)` so it should always parse without errors, as long as the data doesn't get corrupted in transit.

We still need to handle this gracefully, as this can be arbitrary data sent by anyone. Otherwise, with `Parse()` throwing an exception, we open up a potential DoS vector for the master servers. (In fact, the reason we are introducing this feature is that we receive messages from servers that we don't want to receive any from. They happen to have a compatible version format currently, but that can change any moment in theory).

With something like this we should be safe, rough sketch:
```csharp
            if (!Version.TryParse(msgData.ServerVersion, out var serverVersion))
            {
                LunaLog.Debug(...);
            }
            if (!IsValidServerVersion(serverVersion))
```

After fixing that, can you do a rebase of the PR, to get rid of the runtime checker commits that are now merged, and also squash the commits while on it? Thanks!

ShiralynDev (commented):
Won't this ruin with the git history like the other pull request? (The merge branch commit)

BraveCaperCat2:
> Won't this ruin with the git history like the other pull request? (The merge branch commit)

No, the other PR ruined the git history only because I tried to squash the commits together

DasSkelett:
it won't ruin it but it's not a clean history either. @BraveCaperCat2 can you try the following?
1. First squash the commits:
  ```bash
  git rebase -i 3d82ece~1 # this opens an interactive rebase on top of one commit before 3d82ece
  ```
2. In the text in the editor that opens:
  2.1. change the action in the first column for the first commit (`3d82ece`) to `reword` or `r`
  2.2. change the action for the two commits `2b7c7747` (third to last) and `558b5bc1` (second to last) that came through the  `Merge branch 'LunaMultiplayer:master' into feat/server-version-filter` commit to `drop` or `d`
  2.3. change all others to `fixup` or `f`
3. Save and close
4. Now the editor should show the commit message of `3d82ece` (the first commit of this series), and you can change it to something like "Add a server version filter to master servers"
5. Save and close
6. Now you should only have one commit, with all changes of this PR combined, you can confirm this with:
  ```bash
  git log --oneline -4
  ```
  It should only show one commit after `765851ec Update the runtime checker...`, with the new message you gave it.
7. Rebase on top of current master:
  ```bash
  git fetch --all
  git rebase upstream/master
  ```
8. Force-push the new clean branch to GitHub:
  ```bash
  git push --force origin
  ```

BraveCaperCat2:
> it won't ruin it but it's not a clean history either. @BraveCaperCat2 can you try the following?
> 
> 1. First squash the commits:
> 
> ```shell
> git rebase -i 3d82ece~1 # this opens an interactive rebase on top of one commit before 3d82ece
> ```
> 
> 2. In the text in the editor that opens:
>    2.1. change the action in the first column for the first commit (`3d82ece`) to `reword` or `r`
>    2.2. change the action for the two commits `2b7c7747` (third to last) and `558b5bc1` (second to last) that came through the  `Merge branch 'LunaMultiplayer:master' into feat/server-version-filter` commit to `drop` or `d`
>    2.3. change all others to `fixup` or `f`
> 3. Save and close
> 4. Now the editor should show the commit message of `3d82ece` (the first commit of this series), and you can change it to something like "Add a server version filter to master servers"
> 5. Save and close
> 6. Now you should only have one commit, with all changes of this PR combined, you can confirm this with:
> 
> ```shell
> git log --oneline -4
> ```
> 
> It should only show one commit after `765851ec Update the runtime checker...`, with the new message you gave it. 7. Rebase on top of current master:
> 
> ```shell
> git fetch --all
> git rebase upstream/master
> ```
> 
> 8. Force-push the new clean branch to GitHub:
> 
> ```shell
> git push --force origin
> ```

Done.

DasSkelett (approved):

PR #662, opened 2026-04-30
Time Paradoxes Fix

BraveCaperCat2:
Attempt to fix the issue with vessel-related time paradoxes
Fix many of the issues I noticed with the above fix along the way
Modify the code of pretty much every vessel related message handler on the client
Prevent all vessel updates from the future from being applied immediately on the client
Add comments for all of the above

### Additional Notes:
It might be better to rebase and merge or squash and merge this PR (when it's ready, of course) rather than just plain merging it, since there are so many commits.


BraveCaperCat2:
This branch has been tested and has been confirmed to fix the bug without using too much memory.

LmpClient/Systems/VesselActionGroupSys/VesselActionGroupMessageHandler.cs
```diff
@@ -29,5 +50,33 @@ public void HandleMessage(IServerMessageBase msg)
                 queue.Enqueue(msgData);
             }
         }
+
+        // Search through all stored updates and apply any that are no longer from the future
+        public void OnUpdate()
+        {
+            while (StoredMessagesData.TryDequeue(out var StoredMessageData))
```
DrewBanyai: Blocker — OnUpdate will infinite-loop and freeze the game whenever any stored message is still "from the future."

The while (TryDequeue(...)) loop dequeues an item, and if it's still future, re-enqueues it on the same ConcurrentQueue. The very next iteration TryDequeue's it back out, finds it's still future, re-enqueues it, and so on. As long as one item in the queue stays future-dated for the duration of this call, the loop never terminates.

Because this method is wired to TimingManager.UpdateAdd(TimingStage.BetterLateThanNever, MessageHandler.OnUpdate) — i.e. it runs every frame on Unity's main thread — the moment a single from-future message arrives during warp, the game will lock up.

Please bound the iteration by a snapshot of the queue length so each OnUpdate only inspects items that were already in the queue when it started

A better long-term fix is to store messages in a SortedList/min-heap keyed by GameTime and stop on the first one that's still future — then OnUpdate is O(1) in the steady state.

The same bug exists in all 13 new OnUpdate() methods, so please extract this loop into a single shared base class rather than duplicating the fix.

LmpClient/Systems/VesselCoupleSys/VesselCoupleMessageHandler.cs
```diff
@@ -55,5 +78,33 @@ public void HandleMessage(IServerMessageBase msg)
                 }
             }
         }
+
+        // Search through all stored updates and apply any that are no longer from the future
+        public void OnUpdate()
+        {
+            while (StoredMessagesData.TryDequeue(out var StoredMessageData))
```
DrewBanyai: Same infinite-loop bug as in VesselActionGroupMessageHandler.OnUpdate — see my detailed comment there. Please fix in a shared base class so this only has to be addressed once.

LmpClient/Systems/VesselDecoupleSys/VesselDecoupleMessageHandler.cs
```diff
@@ -35,5 +57,33 @@ public void HandleMessage(IServerMessageBase msg)
                 queue.Enqueue(msgData);
             }
         }
+        
+        // Search through all stored updates and apply any that are no longer from the future
+        public void OnUpdate()
+        {
+            while (StoredMessagesData.TryDequeue(out var StoredMessageData))
```
DrewBanyai: Same infinite-loop bug as in VesselActionGroupMessageHandler.OnUpdate — see my detailed comment there. Please fix in a shared base class so this only has to be addressed once.

LmpClient/Systems/VesselFairingsSys/VesselFairingsMessageHandler.cs
```diff
@@ -35,5 +56,33 @@ public void HandleMessage(IServerMessageBase msg)
                 queue.Enqueue(msgData);
             }
         }
+        
+        // Search through all stored updates and apply any that are no longer from the future
+        public void OnUpdate()
+        {
+            while (StoredMessagesData.TryDequeue(out var StoredMessageData))
```
DrewBanyai: Same infinite-loop bug as in VesselActionGroupMessageHandler.OnUpdate — see my detailed comment there. Please fix in a shared base class so this only has to be addressed once.

LmpClient/Systems/VesselFlightStateSys/VesselFlightStateMessageHandler.cs
```diff
@@ -38,5 +60,33 @@ public void HandleMessage(IServerMessageBase msg)
                 queue?.Enqueue(msgData);
             }
         }
+
+        // Search through all stored updates and apply any that are no longer from the future
+        public void OnUpdate()
+        {
+            while (StoredMessagesData.TryDequeue(out var StoredMessageData))
```
DrewBanyai: Same infinite-loop bug as in VesselActionGroupMessageHandler.OnUpdate — see my detailed comment there. Please fix in a shared base class so this only has to be addressed once.

LmpClient/Systems/VesselPartSyncCallSys/VesselPartSyncCallMessageHandler.cs
```diff
@@ -29,5 +50,33 @@ public void HandleMessage(IServerMessageBase msg)
                 queue.Enqueue(msgData);
             }
         }
+
+        // Search through all stored updates and apply any that are no longer from the future
+        public void OnUpdate()
+        {
+            while (StoredMessagesData.TryDequeue(out var StoredMessageData))
```
DrewBanyai: Same infinite-loop bug as in VesselActionGroupMessageHandler.OnUpdate — see my detailed comment there. Please fix in a shared base class so this only has to be addressed once.

LmpClient/Systems/VesselPartSyncFieldSys/VesselPartSyncFieldMessageHandler.cs
```diff
@@ -29,5 +50,33 @@ public void HandleMessage(IServerMessageBase msg)
                 queue.Enqueue(msgData);
             }
         }
+
+        // Search through all stored updates and apply any that are no longer from the future
+        public void OnUpdate()
+        {
+            while (StoredMessagesData.TryDequeue(out var StoredMessageData))
```
DrewBanyai: Same infinite-loop bug as in VesselActionGroupMessageHandler.OnUpdate — see my detailed comment there. Please fix in a shared base class so this only has to be addressed once.

LmpClient/Systems/VesselPartSyncUiFieldSys/VesselPartSyncUiFieldMessageHandler.cs
```diff
@@ -29,5 +50,33 @@ public void HandleMessage(IServerMessageBase msg)
                 queue.Enqueue(msgData);
             }
         }
+
+        // Search through all stored updates and apply any that are no longer from the future
+        public void OnUpdate()
+        {
+            while (StoredMessagesData.TryDequeue(out var StoredMessageData))
```
DrewBanyai: Same infinite-loop bug as in VesselActionGroupMessageHandler.OnUpdate — see my detailed comment there. Please fix in a shared base class so this only has to be addressed once.

LmpClient/Systems/VesselPositionSys/VesselPositionMessageHandler.cs
```diff
@@ -30,5 +56,33 @@ public void HandleMessage(IServerMessageBase msg)
                 queue?.Enqueue(msgData);
             }
         }
+
+        // Search through all stored updates and apply any that are no longer from the future
+        public void OnUpdate()
+        {
+            while (StoredMessagesData.TryDequeue(out var StoredMessageData))
```
DrewBanyai: Same infinite-loop bug as in VesselActionGroupMessageHandler.OnUpdate — see my detailed comment there. Please fix in a shared base class so this only has to be addressed once.

LmpClient/Systems/VesselProtoSys/VesselProtoMessageHandler.cs
```diff
@@ -30,5 +52,33 @@ public void HandleMessage(IServerMessageBase msg)
                 queue.Enqueue(msgData);
             }
         }
+
+        // Search through all stored updates and apply any that are no longer from the future
+        public void OnUpdate()
+        {
+            while (StoredMessagesData.TryDequeue(out var StoredMessageData))
```
DrewBanyai: Same infinite-loop bug as in VesselActionGroupMessageHandler.OnUpdate — see my detailed comment there. Please fix in a shared base class so this only has to be addressed once.

LmpClient/Systems/VesselRemoveSys/VesselRemoveMessageHandler.cs
```diff
@@ -1,24 +1,73 @@
...
+        
+        public void TryQueueUpdate(VesselRemoveMsgData msgData)
+        {
+            System.KillVessel(msgData.VesselId, msgData.AddToKillList, "Applied a vessel remove message from the server");
+        }
+
+        // Search through all stored updates and apply any that are no longer from the future
+        public void OnUpdate()
+        {
+            while (StoredMessagesData.TryDequeue(out var StoredMessageData))
```
DrewBanyai: Same infinite-loop bug as in VesselActionGroupMessageHandler.OnUpdate — see my detailed comment there. Please fix in a shared base class so this only has to be addressed once.

LmpClient/Systems/VesselResourceSys/VesselResourceMessageHandler.cs
```diff
@@ -29,5 +50,33 @@ public void HandleMessage(IServerMessageBase msg)
                 queue.Enqueue(msgData);
             }
         }
+
+        // Search through all stored updates and apply any that are no longer from the future
+        public void OnUpdate()
+        {
+            while (StoredMessagesData.TryDequeue(out var StoredMessageData))
```
DrewBanyai: Same infinite-loop bug as in VesselActionGroupMessageHandler.OnUpdate — see my detailed comment there. Please fix in a shared base class so this only has to be addressed once.

LmpClient/Systems/VesselUndockSys/VesselUndockMessageHandler.cs
```diff
@@ -35,5 +56,33 @@ public void HandleMessage(IServerMessageBase msg)
                 queue.Enqueue(msgData);
             }
         }
+
+        // Search through all stored updates and apply any that are no longer from the future
+        public void OnUpdate()
+        {
+            while (StoredMessagesData.TryDequeue(out var StoredMessageData))
```
DrewBanyai: Same infinite-loop bug as in VesselActionGroupMessageHandler.OnUpdate — see my detailed comment there. Please fix in a shared base class so this only has to be addressed once.

LmpClient/Systems/VesselUpdateSys/VesselUpdateMessageHandler.cs
```diff
@@ -29,5 +50,33 @@ public void HandleMessage(IServerMessageBase msg)
                 queue.Enqueue(msgData);
             }
         }
+
+        // Search through all stored updates and apply any that are no longer from the future
+        public void OnUpdate()
+        {
+            while (StoredMessagesData.TryDequeue(out var StoredMessageData))
```
DrewBanyai: Same infinite-loop bug as in VesselActionGroupMessageHandler.OnUpdate — see my detailed comment there. Please fix in a shared base class so this only has to be addressed once.

LmpClient/VesselUtilities/VesselCommon.cs
```diff
@@ -33,6 +33,13 @@ public static bool UpdateIsForOwnVessel(Guid vesselId)
             return !IsSpectating && FlightGlobals.ActiveVessel && FlightGlobals.ActiveVessel.id == vesselId;
         }
 
+        /// <summary>
+        /// Used to checks if an update came from the (in-game) future.
+        /// It simply checks if the update's timestamp (provided first) is greater than the local timestamp and returns true if it is.
+        /// We might make this function more complex in the future, though.
+        /// </summary>
+        public static bool UpdateIsFromFuture(double UpdateTimestamp, double LocalTimestamp) => UpdateTimestamp > LocalTimestamp;
```
DrewBanyai: Blocker — this predicate is the wrong abstraction for LMP's sync model and will break normal multiplayer.

LMP coordinates clients via subspaces, not raw timestamps. WarpSystem already exposes the proper concepts (SubspaceIsInThePast, SubspaceIsEqualOrInThePast, WarpIfSubspaceIsMoreAdvanced, Subspaces[playerSubspace] < Subspaces[CurrentSubspace], etc.), and VesselPositionUpdate.SubspaceId is already threaded through the position pipeline.

A naive UpdateTimestamp > LocalTimestamp returns true for every packet originating in a more-advanced subspace — which is the normal, expected case any time another player is warped ahead of you. With this predicate driving the new "defer to the future queue" path in 13 subsystems, anyone behind another player will:
- stop seeing other vessels move (VesselPositionMessageHandler),
- stop receiving their control inputs (VesselFlightStateMessageHandler),
- never see vessel deletions (VesselRemoveMessageHandler) — leaving ghost vessels,
- skip the warp-to-coupling-subspace handshake (VesselCoupleMessageHandler — see my inline comment there),
... all while their stored-message queues grow unboundedly until they finally warp forward.

Please drive the "from the future" decision from the existing subspace API instead.

LmpClient/Systems/VesselCoupleSys/VesselCoupleMessageHandler.cs
```diff
@@ -22,16 +27,34 @@ public void HandleMessage(IServerMessageBase msg)
             //Therefore, we must manually call VesselWillBeKilled and implement only 1 of the checks
             if (VesselRemoveSystem.Singleton.VesselWillBeKilled(msgData.VesselId))
                 return;
+            
+            // This code helps prevent time paradoxes.
+            // Why? Because this code makes sure that we only apply updates that aren't from the future.
+            // Additionally, to prevent us from getting desynchronised, we store updates from the future, so that we can apply them later.
+            var IsFromFuture = VesselCommon.UpdateIsFromFuture(msgData.GameTime, WarpSystem.Singleton.CurrentSubspaceTime);
```
DrewBanyai, line 47: Blocker — this enqueue happens before WarpIfSubspaceIsMoreAdvanced, defeating the existing coupling-subspace handshake.

Today, when a coupling packet involves the local active vessel, the client jumps to the future subspace so it stays in sync with whoever performed the dock. With this change, any coupling packet from a more-advanced subspace is short-circuited into StoredMessagesData and returns before WarpSystem.Singleton.WarpIfSubspaceIsMoreAdvanced(msgData.SubspaceId) has a chance to run.

The // Note: This code may not play nicely with the time paradoxes fix... If something breaks with this system when players warp, then this code is probably the cause comment in this PR confirms the regression is known. A known-broken comment isn't a fix — please remove the comment and either:

1. Move the from-future check after the WarpIfSubspaceIsMoreAdvanced(...) call, or
2. Skip the from-future check entirely for couple/decouple/undock and rely on the existing subspace warp-forward behavior, or
3. Trigger the warp first and then process the queued message in OnUpdate once the local subspace catches up.
4. 
Whichever you pick, the active-vessel coupling case must still cause the local client to warp forward.

LmpClient/Systems/VesselRemoveSys/VesselRemoveMessageHandler.cs
```diff
@@ -1,24 +1,73 @@
...
             if (!VesselCommon.IsSpectating && FlightGlobals.ActiveVessel && FlightGlobals.ActiveVessel.id == msgData.VesselId)
                 return;
+            
+            // This code helps prevent time paradoxes.
+            // Why? Because this code makes sure that we only apply updates that aren't from the future.
+            // Additionally, to prevent us from getting desynchronised, we store updates from the future, so that we can apply them later.
+            var IsFromFuture = VesselCommon.UpdateIsFromFuture(msgData.GameTime, WarpSystem.Singleton.CurrentSubspaceTime);
 
-            System.KillVessel(msgData.VesselId, msgData.AddToKillList, "Received a vessel remove message from server");
+            if (IsFromFuture)
```
DrewBanyai, line 34: Blocker — deferring vessel removals based on GameTime will leave ghost vessels in the world.

Vessel removal isn't a positional update that can be safely interpolated or replayed later — it's a destructive event the server has already committed in its (possibly more-advanced) subspace. By enqueueing remove messages from any "future" subspace into StoredMessagesData, the client keeps simulating, rendering, and potentially colliding with vessels that no longer exist server-side, until the local clock finally catches up. In a session where another player is warped ahead, that may be never.

Combined with the infinite-loop bug in OnUpdate (see my comment there), a single deferred remove can also wedge the main thread.

Please either:
- skip the from-future check entirely for VesselRemove (apply removals immediately — they're authoritative), or
- apply the removal immediately and warp the local player to msgData.SubspaceId so the rest of their state stays consistent.

Side note: TryQueueUpdate here just calls KillVessel with a slightly different log string ("Applied..." vs "Received..."). Once the from-future branch is reworked, please collapse the two paths into one helper so they can't drift.

DrewBanyai (changes requested):

LmpClient/Systems/VesselRemoveSys/VesselRemoveMessageHandler.cs
BraveCaperCat2: Removed vessels would probably be removed because something happened to them. Those vessels should still exist until that event had happened locally. "Simulating, rendering and potentially colliding with vessels that no longer exist server-side" will only happen until the local clock catches up to the point where the vessel was originally removed, not until it catches up with the player who sent it.

Calling them "ghost vessels" however is completely warranted.

LmpClient/Systems/VesselCoupleSys/VesselCoupleMessageHandler.cs
BraveCaperCat2: I had not understood the purpose of the `WarpIfSubspaceIsMoreAdvanced()` call when I was writing that code. It seemed to serve a similar purpose (albeit much more jarring) to the code I was adding to the file, as it *also* checked whether the message was from the future (albeit hidden behind the `WarpIfSubspaceIsMoreAdvanced` call) and took action to prevent time paradoxes. (warping to the other subspace) My code, however, would wait for the docking vessel to arrive before docking it, rather than just warping to the future.

It got me really confused, especially after no one had responded to me specifically asking whether I should put my code below or above that code.

I ended up writing the comment because I was still pretty unsure where my code should have been in the file at the time. I did not say that it was "known to be broken", I said that it "may cause issues".

LmpClient/VesselUtilities/VesselCommon.cs
BraveCaperCat2: Raw timestamps are necessary. Say that you're a minute or so behind another player (quite unrealistic, but sure) and they make a burn to escape the mun. You should only have to be at a time after the burn was completed to be able to see the results, not in a subspace after the one that completed the burn. The difference is that if you never warp for whatever reason, your subspace will never surpass that of the other player, where-as with the current behaviour, you'd just have to wait about a minute and you'd see the burn.

Even worse, imagine that the other player left the game and their subspace was destroyed by the server - we'd error constantly because that player's subspace no longer exists, causing severe lag on the client. Either that, or we'd have to drop the messages entirely, which would cause the client to get permanently desynchronised from the server.

As for the behaviours of `stop seeing other vessels move` and `stop receiving their control inputs`, you appear to be misinterpreting what my code does. (again) The messages sent by other players will be played back at the same game timestamps they were sent at, so you'd still see other players move - you'd just be "lagged" behind them until you warped to catch up. Vessel deletions and warp-to-coupling-subspace is more difficult to explain away, please see my other replies for that.

LmpClient/Systems/VesselActionGroupSys/VesselActionGroupMessageHandler.cs
BraveCaperCat2: This would appear to be a very serious issue - I thought that just naively iterating through the queue with an index would not work because we'd be modifying the queue (dequeuing items) while iterating through the queue. Programming in the lua language has engraved into my mind that I should never modify a list while iterating through it, which has probably come back to haunt me here.

Perhaps more surprising is that I didn't encounter this in testing. The client that was in the past should have instantly frozen as soon as the other client warped, but it didn't. You can see this in the test footage I provided - no freezing in sight.

LmpClient/Systems/VesselCoupleSys/VesselCoupleMessageHandler.cs
BraveCaperCat2: See my other reply.

LmpClient/Systems/VesselDecoupleSys/VesselDecoupleMessageHandler.cs
BraveCaperCat2: See my other reply.

LmpClient/Systems/VesselFairingsSys/VesselFairingsMessageHandler.cs
BraveCaperCat2: See my other reply.

LmpClient/Systems/VesselFlightStateSys/VesselFlightStateMessageHandler.cs
BraveCaperCat2: See my other reply.

LmpClient/Systems/VesselPartSyncCallSys/VesselPartSyncCallMessageHandler.cs
BraveCaperCat2: See my other reply.

LmpClient/Systems/VesselPartSyncFieldSys/VesselPartSyncFieldMessageHandler.cs
BraveCaperCat2: See my other reply.

LmpClient/Systems/VesselPartSyncUiFieldSys/VesselPartSyncUiFieldMessageHandler.cs
BraveCaperCat2: See my other reply.

LmpClient/Systems/VesselPositionSys/VesselPositionMessageHandler.cs
BraveCaperCat2: See my other reply.

LmpClient/Systems/VesselProtoSys/VesselProtoMessageHandler.cs
BraveCaperCat2: See my other reply.

LmpClient/Systems/VesselRemoveSys/VesselRemoveMessageHandler.cs
BraveCaperCat2: See my other reply.

LmpClient/Systems/VesselResourceSys/VesselResourceMessageHandler.cs
BraveCaperCat2: See my other reply.

LmpClient/Systems/VesselUndockSys/VesselUndockMessageHandler.cs
BraveCaperCat2: See my other reply.

LmpClient/Systems/VesselUpdateSys/VesselUpdateMessageHandler.cs
BraveCaperCat2: See my other reply.

LmpClient/Systems/VesselActionGroupSys/VesselActionGroupMessageHandler.cs
BraveCaperCat2: > please extract this loop into a single shared base class rather than duplicating the fix.

I'm not exactly sure how to do this... do I just make a `VesselBaseMessageHandler`, give that the fix and make everything else inherit from it? What about the `// This code helps prevent time paradoxes.` code? Do I put that in `VesselBaseMessageHandler` too? What about `VesselPositionMessageHandler` and `VesselCoupleMessageHandler` which need special treatment?

LmpClient/Systems/VesselCoupleSys/VesselCoupleMessageHandler.cs
BraveCaperCat2: Additionally, the decouple and undock systems didn't have similar code and aren't affected by this particular issue.

BraveCaperCat2: This is fixed with [commit 76d6bae](https://github.com/LunaMultiplayer/LunaMultiplayer/pull/662/commits/76d6bae95f238164861e3dad52de41f344e47f99).

gavazquez:
Wouldn't make more sense that the server relays the messages to the correct clients? 

I think that it already contains a dictionary of the players and the current subspaces. So it's a matter of relaying the messages to the clients that are past the message time and store the future ones.

Just an idea ;)

BraveCaperCat2:
> Wouldn't make more sense that the server relays the messages to the correct clients?
> 
> I think that it already contains a dictionary of the players and the current subspaces. So it's a matter of relaying the messages to the clients that are past the message time and store the future ones.
> 
> Just an idea ;)

I'm not sure how that's any better than the solution we're currently working on

LmpClient/Systems/VesselActionGroupSys/VesselActionGroupMessageHandler.cs
BraveCaperCat2: This issue should be fixed in the latest commit

LmpClient/Systems/VesselResourceSys/VesselResourceMessageHandler.cs
```diff
@@ -29,5 +66,43 @@ public void HandleMessage(IServerMessageBase msg)
                 queue.Enqueue(msgData);
             }
         }
+
+        // Search through all stored updates and apply any that are no longer from the future
+        public void OnUpdate()
```
DrewBanyai, line 75: There are 13 handlers of this type and each one has the same issue: The cleanup condition is inverted. Pass messages will be replayed every frame and never removed. Think about it this way: Whenevere it something to remove, you return out and the removal is what would occur right after if you had not returned... This will cause them to be processed over and over and over again.

LmpClient/Systems/VesselCoupleSys/VesselCoupleMessageHandler.cs
```diff
@@ -55,5 +93,43 @@ public void HandleMessage(IServerMessageBase msg)
                 }
             }
         }
+
+        // Search through all stored updates and apply any that are no longer from the future
+        public void OnUpdate()
```
DrewBanyai, line 102: There are 13 handlers of this type and each one has the same issue: The cleanup condition is inverted. Pass messages will be replayed every frame and never removed. Think about it this way: Whenevere it something to remove, you return out and the removal is what would occur right after if you had not returned... This will cause them to be processed over and over and over again.

LmpClient/Systems/VesselDecoupleSys/VesselDecoupleMessageHandler.cs
```diff
@@ -35,5 +73,43 @@ public void HandleMessage(IServerMessageBase msg)
                 queue.Enqueue(msgData);
             }
         }
+
+        // Search through all stored updates and apply any that are no longer from the future
+        public void OnUpdate()
+        {
+            var softDeleted = new List<double>();
+            foreach (var kvp in StoredMessagesData)
```
DrewBanyai: There are 13 handlers of this type and each one has the same issue: The cleanup condition is inverted. Pass messages will be replayed every frame and never removed. Think about it this way: Whenevere it something to remove, you return out and the removal is what would occur right after if you had not returned... This will cause them to be processed over and over and over again.

LmpClient/Systems/VesselFairingsSys/VesselFairingsMessageHandler.cs
```diff
@@ -35,5 +72,43 @@ public void HandleMessage(IServerMessageBase msg)
                 queue.Enqueue(msgData);
             }
         }
+
+        // Search through all stored updates and apply any that are no longer from the future
+        public void OnUpdate()
+        {
+            var softDeleted = new List<double>();
+            foreach (var kvp in StoredMessagesData)
```
DrewBanyai: There are 13 handlers of this type and each one has the same issue: The cleanup condition is inverted. Pass messages will be replayed every frame and never removed. Think about it this way: Whenevere it something to remove, you return out and the removal is what would occur right after if you had not returned... This will cause them to be processed over and over and over again.

LmpClient/Systems/VesselFlightStateSys/VesselFlightStateMessageHandler.cs
```diff
@@ -38,5 +76,43 @@ public void HandleMessage(IServerMessageBase msg)
                 queue?.Enqueue(msgData);
             }
         }
+
+        // Search through all stored updates and apply any that are no longer from the future
+        public void OnUpdate()
+        {
+            var softDeleted = new List<double>();
+            foreach (var kvp in StoredMessagesData)
```
DrewBanyai: There are 13 handlers of this type and each one has the same issue: The cleanup condition is inverted. Pass messages will be replayed every frame and never removed. Think about it this way: Whenevere it something to remove, you return out and the removal is what would occur right after if you had not returned... This will cause them to be processed over and over and over again.

LmpClient/Systems/VesselPartSyncCallSys/VesselPartSyncCallMessageHandler.cs
```diff
@@ -29,5 +66,43 @@ public void HandleMessage(IServerMessageBase msg)
                 queue.Enqueue(msgData);
             }
         }
+
+        // Search through all stored updates and apply any that are no longer from the future
+        public void OnUpdate()
+        {
+            var softDeleted = new List<double>();
+            foreach (var kvp in StoredMessagesData)
```
DrewBanyai: There are 13 handlers of this type and each one has the same issue: The cleanup condition is inverted. Pass messages will be replayed every frame and never removed. Think about it this way: Whenevere it something to remove, you return out and the removal is what would occur right after if you had not returned... This will cause them to be processed over and over and over again.

LmpClient/Systems/VesselPartSyncFieldSys/VesselPartSyncFieldMessageHandler.cs
```diff
@@ -29,5 +66,43 @@ public void HandleMessage(IServerMessageBase msg)
                 queue.Enqueue(msgData);
             }
         }
+
+        // Search through all stored updates and apply any that are no longer from the future
+        public void OnUpdate()
```
DrewBanyai, line 75: There are 13 handlers of this type and each one has the same issue: The cleanup condition is inverted. Pass messages will be replayed every frame and never removed. Think about it this way: Whenevere it something to remove, you return out and the removal is what would occur right after if you had not returned... This will cause them to be processed over and over and over again.

LmpClient/Systems/VesselPartSyncUiFieldSys/VesselPartSyncUiFieldMessageHandler.cs
```diff
@@ -29,5 +66,43 @@ public void HandleMessage(IServerMessageBase msg)
                 queue.Enqueue(msgData);
             }
         }
+
+        // Search through all stored updates and apply any that are no longer from the future
+        public void OnUpdate()
```
DrewBanyai, line 75: There are 13 handlers of this type and each one has the same issue: The cleanup condition is inverted. Pass messages will be replayed every frame and never removed. Think about it this way: Whenevere it something to remove, you return out and the removal is what would occur right after if you had not returned... This will cause them to be processed over and over and over again.

LmpClient/Systems/VesselPositionSys/VesselPositionMessageHandler.cs
```diff
@@ -30,5 +72,43 @@ public void HandleMessage(IServerMessageBase msg)
                 queue?.Enqueue(msgData);
             }
         }
+
+        // Search through all stored updates and apply any that are no longer from the future
+        public void OnUpdate()
```
DrewBanyai, line 81: There are 13 handlers of this type and each one has the same issue: The cleanup condition is inverted. Pass messages will be replayed every frame and never removed. Think about it this way: Whenevere it something to remove, you return out and the removal is what would occur right after if you had not returned... This will cause them to be processed over and over and over again.

LmpClient/Systems/VesselProtoSys/VesselProtoMessageHandler.cs
```diff
@@ -30,5 +68,43 @@ public void HandleMessage(IServerMessageBase msg)
                 queue.Enqueue(msgData);
             }
         }
+
+        // Search through all stored updates and apply any that are no longer from the future
+        public void OnUpdate()
```
DrewBanyai, line 77: There are 13 handlers of this type and each one has the same issue: The cleanup condition is inverted. Pass messages will be replayed every frame and never removed. Think about it this way: Whenevere it something to remove, you return out and the removal is what would occur right after if you had not returned... This will cause them to be processed over and over and over again.

LmpClient/Systems/VesselRemoveSys/VesselRemoveMessageHandler.cs
```diff
@@ -1,24 +1,99 @@
...
+            System.KillVessel(msgData.VesselId, msgData.AddToKillList, "Received a vessel remove message from the server");
+        }
+        
+        public void TryQueueUpdate(VesselRemoveMsgData msgData)
+        {
+            System.KillVessel(msgData.VesselId, msgData.AddToKillList, "Applied a vessel remove message from the server");
+        }
+
+        // Search through all stored updates and apply any that are no longer from the future
+        public void OnUpdate()
```
DrewBanyai, line 66: There are 13 handlers of this type and each one has the same issue: The cleanup condition is inverted. Pass messages will be replayed every frame and never removed. Think about it this way: Whenevere it something to remove, you return out and the removal is what would occur right after if you had not returned... This will cause them to be processed over and over and over again.

LmpClient/Systems/VesselUndockSys/VesselUndockMessageHandler.cs
```diff
@@ -35,5 +72,43 @@ public void HandleMessage(IServerMessageBase msg)
                 queue.Enqueue(msgData);
             }
         }
+
+        // Search through all stored updates and apply any that are no longer from the future
+        public void OnUpdate()
```
DrewBanyai, line 81: There are 13 handlers of this type and each one has the same issue: The cleanup condition is inverted. Pass messages will be replayed every frame and never removed. Think about it this way: Whenevere it something to remove, you return out and the removal is what would occur right after if you had not returned... This will cause them to be processed over and over and over again.

LmpClient/Systems/VesselUpdateSys/VesselUpdateMessageHandler.cs
```diff
@@ -29,5 +65,43 @@ public void HandleMessage(IServerMessageBase msg)
                 queue.Enqueue(msgData);
             }
         }
+
+        // Search through all stored updates and apply any that are no longer from the future
+        public void OnUpdate()
```
DrewBanyai, line 75: There are 13 handlers of this type and each one has the same issue: The cleanup condition is inverted. Pass messages will be replayed every frame and never removed. Think about it this way: Whenevere it something to remove, you return out and the removal is what would occur right after if you had not returned... This will cause them to be processed over and over and over again.

LmpClient/Systems/VesselActionGroupSys/VesselActionGroupMessageHandler.cs
```diff
@@ -29,5 +66,43 @@ public void HandleMessage(IServerMessageBase msg)
                 queue.Enqueue(msgData);
             }
         }
+
+        // Search through all stored updates and apply any that are no longer from the future
+        public void OnUpdate()
```
DrewBanyai, line 75: There are 13 handlers of this type and each one has the same issue: The cleanup condition is inverted. Pass messages will be replayed every frame and never removed. Think about it this way: Whenevere it something to remove, you return out and the removal is what would occur right after if you had not returned... This will cause them to be processed over and over and over again.

LmpClient/Systems/VesselPartSyncUiFieldSys/VesselPartSyncUiFieldMessageHandler.cs
```diff
@@ -1,15 +1,21 @@
...
+using System;
+using System.Linq;
 
 namespace LmpClient.Systems.VesselPartSyncUiFieldSys
 {
     public class VesselPartSyncUiFieldMessageHandler : SubSystem<VesselPartSyncUiFieldSystem>, IMessageHandler
     {
         public ConcurrentQueue<IServerMessageBase> IncomingMessages { get; set; } = new ConcurrentQueue<IServerMessageBase>();
+        
+        public SortedList<double,VesselPartSyncUiFieldMsgData> StoredMessagesData { get; set; } = new SortedList<double,VesselPartSyncUiFieldMsgData>();
```
DrewBanyai, line 18: This is a great addition in terms of replacing the ConcurrentQueue + while from the previous version, but you've stumbled into another blocking issue with it. This isn't thread-safe, because you're accessing it from both a HandleMessage call (which is on a background task) and OnUpdate (which is on the main Unity thread).

This will cause it to either drop InvalidOperationException, IndexOutOfRangeExceptions, or worse yet, potentially silently correupted internal arrays. ConcurrentQueue was specifically previously used to prevent this threading issue, if I had to guess. So this is a regression unless you utilize locks to ensure thread safety.

LmpClient/VesselUtilities/VesselCommon.cs
```diff
@@ -126,6 +133,25 @@ public static bool DoVesselChecks(Guid vesselId)
...
         }
 
+        /// <summary>
+        /// Checks if the vessel with the provided ID has the provided orbit and returns true if it does.
+        /// </summary>
+        public static bool VesselHasSameOrbit(Guid vesselId, double[] orbit)
+        {
+            var vessel = FlightGlobals.FindVessel(vesselId);
+
+            var delta = Math.Pow(10, -15); // Acceptable margin of error
```
DrewBanyai: This is such a small margin that basically any semMajorAxis value from 10^6 to 10^9 will always return false on it, which makes the VesselPositionMessageHandler "same orbit" fast-path dead code. On top of that, you need to check against the orbit.Length being 0 and the FindVessel / vessel.orbit being null.

LmpClient/Systems/VesselUpdateSys/VesselUpdate.cs
```diff
@@ -30,8 +30,8 @@ public class VesselUpdate
         public string AutoCleanReason;
         public bool WasControllable;
         public int Stage;
-        public string BodyName;
         public float[] Com = new float[3];
+        public string BodyName;
```
DrewBanyai, line 34: I'm not sure if this just got added because you were pulling from VesselUpdateSys's VesselUpdate but in this case it's never written to or read.

LmpClient/Systems/VesselPartSyncUiFieldSys/VesselPartSyncUiFieldMessageHandler.cs
```diff
@@ -29,5 +66,43 @@ public void HandleMessage(IServerMessageBase msg)
...
+            if (softDeleted.Count > 0)
+                return;
+
+            // Finish deleting everything
+            foreach (double key in softDeleted)
+                StoredMessagesData.Remove(key);
+        }
+        
+        // Log out the amount of memory we're using to store messages
+        public void LogQueuedMessagesSize()
```
DrewBanyai, line 108: Could you gate this behind a debug flag so it only shows up in debug builds? As of now, all 13 systems will add noise to the log from this indefinitely.

LmpClient/Systems/VesselActionGroupSys/VesselActionGroupMessageHandler.cs
```diff
@@ -29,5 +66,43 @@ public void HandleMessage(IServerMessageBase msg)
...
+            if (softDeleted.Count > 0)
+                return;
+
+            // Finish deleting everything
+            foreach (double key in softDeleted)
+                StoredMessagesData.Remove(key);
+        }
+
+        // Log out the amount of memory we're using to store messages
+        public void LogQueuedMessagesSize()
```
DrewBanyai, line 108: Could you gate this behind a debug flag so it only shows up in debug builds? As of now, all 13 systems will add noise to the log from this indefinitely.

LmpClient/Systems/VesselCoupleSys/VesselCoupleMessageHandler.cs
```diff
@@ -55,5 +93,43 @@ public void HandleMessage(IServerMessageBase msg)
...
+            if (softDeleted.Count > 0)
+                return;
+
+            // Finish deleting everything
+            foreach (double key in softDeleted)
+                StoredMessagesData.Remove(key);
+        }
+        
+        // Log out the amount of memory we're using to store messages
+        public void LogQueuedMessagesSize()
```
DrewBanyai, line 135: Could you gate this behind a debug flag so it only shows up in debug builds? As of now, all 13 systems will add noise to the log from this indefinitely.

LmpClient/Systems/VesselDecoupleSys/VesselDecoupleMessageHandler.cs
```diff
@@ -35,5 +73,43 @@ public void HandleMessage(IServerMessageBase msg)
...
+            if (softDeleted.Count > 0)
+                return;
+
+            // Finish deleting everything
+            foreach (double key in softDeleted)
+                StoredMessagesData.Remove(key);
+        }
+        
+        // Log out the amount of memory we're using to store messages
+        public void LogQueuedMessagesSize()
```
DrewBanyai, line 115: Could you gate this behind a debug flag so it only shows up in debug builds? As of now, all 13 systems will add noise to the log from this indefinitely.

LmpClient/Systems/VesselFairingsSys/VesselFairingsMessageHandler.cs
```diff
@@ -35,5 +72,43 @@ public void HandleMessage(IServerMessageBase msg)
...
+            if (softDeleted.Count > 0)
+                return;
+
+            // Finish deleting everything
+            foreach (double key in softDeleted)
+                StoredMessagesData.Remove(key);
+        }
+        
+        // Log out the amount of memory we're using to store messages
+        public void LogQueuedMessagesSize()
```
DrewBanyai, line 114: Could you gate this behind a debug flag so it only shows up in debug builds? As of now, all 13 systems will add noise to the log from this indefinitely.

LmpClient/Systems/VesselFlightStateSys/VesselFlightStateMessageHandler.cs
```diff
@@ -38,5 +76,43 @@ public void HandleMessage(IServerMessageBase msg)
...
+            if (softDeleted.Count > 0)
+                return;
+
+            // Finish deleting everything
+            foreach (double key in softDeleted)
+                StoredMessagesData.Remove(key);
+        }
+        
+        // Log out the amount of memory we're using to store messages
+        public void LogQueuedMessagesSize()
```
DrewBanyai, line 118: Could you gate this behind a debug flag so it only shows up in debug builds? As of now, all 13 systems will add noise to the log from this indefinitely.

LmpClient/Systems/VesselPartSyncCallSys/VesselPartSyncCallMessageHandler.cs
```diff
@@ -29,5 +66,43 @@ public void HandleMessage(IServerMessageBase msg)
...
+            if (softDeleted.Count > 0)
+                return;
+
+            // Finish deleting everything
+            foreach (double key in softDeleted)
+                StoredMessagesData.Remove(key);
+        }
+        
+        // Log out the amount of memory we're using to store messages
+        public void LogQueuedMessagesSize()
```
DrewBanyai, line 108: Could you gate this behind a debug flag so it only shows up in debug builds? As of now, all 13 systems will add noise to the log from this indefinitely.

LmpClient/Systems/VesselPartSyncFieldSys/VesselPartSyncFieldMessageHandler.cs
```diff
@@ -29,5 +66,43 @@ public void HandleMessage(IServerMessageBase msg)
...
+            if (softDeleted.Count > 0)
+                return;
+
+            // Finish deleting everything
+            foreach (double key in softDeleted)
+                StoredMessagesData.Remove(key);
+        }
+        
+        // Log out the amount of memory we're using to store messages
+        public void LogQueuedMessagesSize()
```
DrewBanyai, line 108: Could you gate this behind a debug flag so it only shows up in debug builds? As of now, all 13 systems will add noise to the log from this indefinitely.

LmpClient/Systems/VesselPositionSys/VesselPositionMessageHandler.cs
```diff
@@ -30,5 +72,43 @@ public void HandleMessage(IServerMessageBase msg)
...
+            if (softDeleted.Count > 0)
+                return;
+
+            // Finish deleting everything
+            foreach (double key in softDeleted)
+                StoredMessagesData.Remove(key);
+        }
+        
+        // Log out the amount of memory we're using to store messages
+        public void LogQueuedMessagesSize()
```
DrewBanyai, line 114: Could you gate this behind a debug flag so it only shows up in debug builds? As of now, all 13 systems will add noise to the log from this indefinitely.

LmpClient/Systems/VesselProtoSys/VesselProtoMessageHandler.cs
```diff
@@ -30,5 +68,43 @@ public void HandleMessage(IServerMessageBase msg)
...
+            if (softDeleted.Count > 0)
+                return;
+
+            // Finish deleting everything
+            foreach (double key in softDeleted)
+                StoredMessagesData.Remove(key);
+        }
+        
+        // Log out the amount of memory we're using to store messages
+        public void LogQueuedMessagesSize()
```
DrewBanyai, line 110: Could you gate this behind a debug flag so it only shows up in debug builds? As of now, all 13 systems will add noise to the log from this indefinitely.

LmpClient/Systems/VesselRemoveSys/VesselRemoveMessageHandler.cs
```diff
@@ -1,24 +1,99 @@
...
+            if (softDeleted.Count > 0)
+                return;
+
+            // Finish deleting everything
+            foreach (double key in softDeleted)
+                StoredMessagesData.Remove(key);
+        }
+        
+        // Log out the amount of memory we're using to store messages
+        public void LogQueuedMessagesSize()
```
DrewBanyai, line 99: Could you gate this behind a debug flag so it only shows up in debug builds? As of now, all 13 systems will add noise to the log from this indefinitely.

LmpClient/Systems/VesselResourceSys/VesselResourceMessageHandler.cs
```diff
@@ -29,5 +66,43 @@ public void HandleMessage(IServerMessageBase msg)
...
+            if (softDeleted.Count > 0)
+                return;
+
+            // Finish deleting everything
+            foreach (double key in softDeleted)
+                StoredMessagesData.Remove(key);
+        }
+        
+        // Log out the amount of memory we're using to store messages
+        public void LogQueuedMessagesSize()
```
DrewBanyai, line 108: Could you gate this behind a debug flag so it only shows up in debug builds? As of now, all 13 systems will add noise to the log from this indefinitely.

LmpClient/Systems/VesselUndockSys/VesselUndockMessageHandler.cs
```diff
@@ -35,5 +72,43 @@ public void HandleMessage(IServerMessageBase msg)
...
+            if (softDeleted.Count > 0)
+                return;
+
+            // Finish deleting everything
+            foreach (double key in softDeleted)
+                StoredMessagesData.Remove(key);
+        }
+        
+        // Log out the amount of memory we're using to store messages
+        public void LogQueuedMessagesSize()
```
DrewBanyai, line 114: Could you gate this behind a debug flag so it only shows up in debug builds? As of now, all 13 systems will add noise to the log from this indefinitely.

LmpClient/Systems/VesselUpdateSys/VesselUpdateMessageHandler.cs
```diff
@@ -29,5 +65,43 @@ public void HandleMessage(IServerMessageBase msg)
...
+            if (softDeleted.Count > 0)
+                return;
+
+            // Finish deleting everything
+            foreach (double key in softDeleted)
+                StoredMessagesData.Remove(key);
+        }
+        
+        // Log out the amount of memory we're using to store messages
+        public void LogQueuedMessagesSize()
```
DrewBanyai, line 108: Could you gate this behind a debug flag so it only shows up in debug builds? As of now, all 13 systems will add noise to the log from this indefinitely.

LmpClient/Systems/VesselCoupleSys/VesselCoupleMessageHandler.cs
```diff
@@ -55,5 +93,43 @@ public void HandleMessage(IServerMessageBase msg)
...
+        {
+            var memUsage = 0;
+            if (StoredMessagesData.Count > 0)
+            {
+                memUsage = StoredMessagesData.Count * StoredMessagesData.First().Value.GetMessageSize() / 1024; // This is in Kilobytes
```
DrewBanyai: This is an int, so dividing by 1024 means you'll get 0 until the value is at or above 1024, which is a lot of bytes to represent as 0 over the course of all the times you're going to write this debug message.

LmpClient/Systems/VesselFairingsSys/VesselFairingsMessageHandler.cs
```diff
@@ -35,5 +72,43 @@ public void HandleMessage(IServerMessageBase msg)
...
+        {
+            var memUsage = 0;
+            if (StoredMessagesData.Count > 0)
+            {
+                memUsage = StoredMessagesData.Count * StoredMessagesData.First().Value.GetMessageSize() / 1024; // This is in Kilobytes
```
DrewBanyai: This is an int, so dividing by 1024 means you'll get 0 until the value is at or above 1024, which is a lot of bytes to represent as 0 over the course of all the times you're going to write this debug message.

LmpClient/Systems/VesselFlightStateSys/VesselFlightStateMessageHandler.cs
```diff
@@ -38,5 +76,43 @@ public void HandleMessage(IServerMessageBase msg)
...
+        {
+            var memUsage = 0;
+            if (StoredMessagesData.Count > 0)
+            {
+                memUsage = StoredMessagesData.Count * StoredMessagesData.First().Value.GetMessageSize() / 1024; // This is in Kilobytes
```
DrewBanyai: This is an int, so dividing by 1024 means you'll get 0 until the value is at or above 1024, which is a lot of bytes to represent as 0 over the course of all the times you're going to write this debug message.

LmpClient/Systems/VesselPartSyncCallSys/VesselPartSyncCallMessageHandler.cs
```diff
@@ -29,5 +66,43 @@ public void HandleMessage(IServerMessageBase msg)
...
+        {
+            var memUsage = 0;
+            if (StoredMessagesData.Count > 0)
+            {
+                memUsage = StoredMessagesData.Count * StoredMessagesData.First().Value.GetMessageSize() / 1024; // This is in Kilobytes
```
DrewBanyai: This is an int, so dividing by 1024 means you'll get 0 until the value is at or above 1024, which is a lot of bytes to represent as 0 over the course of all the times you're going to write this debug message.

LmpClient/Systems/VesselPartSyncFieldSys/VesselPartSyncFieldMessageHandler.cs
```diff
@@ -29,5 +66,43 @@ public void HandleMessage(IServerMessageBase msg)
...
+        {
+            var memUsage = 0;
+            if (StoredMessagesData.Count > 0)
+            {
+                memUsage = StoredMessagesData.Count * StoredMessagesData.First().Value.GetMessageSize() / 1024; // This is in Kilobytes
```
DrewBanyai: This is an int, so dividing by 1024 means you'll get 0 until the value is at or above 1024, which is a lot of bytes to represent as 0 over the course of all the times you're going to write this debug message.

LmpClient/Systems/VesselPartSyncUiFieldSys/VesselPartSyncUiFieldMessageHandler.cs
```diff
@@ -29,5 +66,43 @@ public void HandleMessage(IServerMessageBase msg)
...
+        {
+            var memUsage = 0;
+            if (StoredMessagesData.Count > 0)
+            {
+                memUsage = StoredMessagesData.Count * StoredMessagesData.First().Value.GetMessageSize() / 1024; // This is in Kilobytes
```
DrewBanyai: This is an int, so dividing by 1024 means you'll get 0 until the value is at or above 1024, which is a lot of bytes to represent as 0 over the course of all the times you're going to write this debug message.

LmpClient/Systems/VesselPositionSys/VesselPositionMessageHandler.cs
```diff
@@ -30,5 +72,43 @@ public void HandleMessage(IServerMessageBase msg)
...
+        {
+            var memUsage = 0;
+            if (StoredMessagesData.Count > 0)
+            {
+                memUsage = StoredMessagesData.Count * StoredMessagesData.First().Value.GetMessageSize() / 1024; // This is in Kilobytes
```
DrewBanyai: This is an int, so dividing by 1024 means you'll get 0 until the value is at or above 1024, which is a lot of bytes to represent as 0 over the course of all the times you're going to write this debug message.

LmpClient/Systems/VesselProtoSys/VesselProtoMessageHandler.cs
```diff
@@ -30,5 +68,43 @@ public void HandleMessage(IServerMessageBase msg)
...
+        {
+            var memUsage = 0;
+            if (StoredMessagesData.Count > 0)
+            {
+                memUsage = StoredMessagesData.Count * StoredMessagesData.First().Value.GetMessageSize() / 1024; // This is in Kilobytes
```
DrewBanyai: This is an int, so dividing by 1024 means you'll get 0 until the value is at or above 1024, which is a lot of bytes to represent as 0 over the course of all the times you're going to write this debug message.

LmpClient/Systems/VesselRemoveSys/VesselRemoveMessageHandler.cs
```diff
@@ -1,24 +1,99 @@
...
+        {
+            var memUsage = 0;
+            if (StoredMessagesData.Count > 0)
+            {
+                memUsage = StoredMessagesData.Count * StoredMessagesData.First().Value.GetMessageSize() / 1024; // This is in Kilobytes
```
DrewBanyai: This is an int, so dividing by 1024 means you'll get 0 until the value is at or above 1024, which is a lot of bytes to represent as 0 over the course of all the times you're going to write this debug message.

LmpClient/Systems/VesselResourceSys/VesselResourceMessageHandler.cs
```diff
@@ -29,5 +66,43 @@ public void HandleMessage(IServerMessageBase msg)
...
+        {
+            var memUsage = 0;
+            if (StoredMessagesData.Count > 0)
+            {
+                memUsage = StoredMessagesData.Count * StoredMessagesData.First().Value.GetMessageSize() / 1024; // This is in Kilobytes
```
DrewBanyai: This is an int, so dividing by 1024 means you'll get 0 until the value is at or above 1024, which is a lot of bytes to represent as 0 over the course of all the times you're going to write this debug message.

LmpClient/Systems/VesselUndockSys/VesselUndockMessageHandler.cs
```diff
@@ -35,5 +72,43 @@ public void HandleMessage(IServerMessageBase msg)
...
+        {
+            var memUsage = 0;
+            if (StoredMessagesData.Count > 0)
+            {
+                memUsage = StoredMessagesData.Count * StoredMessagesData.First().Value.GetMessageSize() / 1024; // This is in Kilobytes
```
DrewBanyai: This is an int, so dividing by 1024 means you'll get 0 until the value is at or above 1024, which is a lot of bytes to represent as 0 over the course of all the times you're going to write this debug message.

LmpClient/Systems/VesselUpdateSys/VesselUpdateMessageHandler.cs
```diff
@@ -29,5 +65,43 @@ public void HandleMessage(IServerMessageBase msg)
...
+        {
+            var memUsage = 0;
+            if (StoredMessagesData.Count > 0)
+            {
+                memUsage = StoredMessagesData.Count * StoredMessagesData.First().Value.GetMessageSize() / 1024; // This is in Kilobytes
```
DrewBanyai: This is an int, so dividing by 1024 means you'll get 0 until the value is at or above 1024, which is a lot of bytes to represent as 0 over the course of all the times you're going to write this debug message.

LmpClient/Systems/VesselActionGroupSys/VesselActionGroupMessageHandler.cs
```diff
@@ -29,5 +66,43 @@ public void HandleMessage(IServerMessageBase msg)
...
+        {
+            var memUsage = 0;
+            if (StoredMessagesData.Count > 0)
+            {
+                memUsage = StoredMessagesData.Count * StoredMessagesData.First().Value.GetMessageSize() / 1024; // This is in Kilobytes
```
DrewBanyai: This is an int, so dividing by 1024 means you'll get 0 until the value is at or above 1024, which is a lot of bytes to represent as 0 over the course of all the times you're going to write this debug message.

LmpClient/Systems/VesselDecoupleSys/VesselDecoupleMessageHandler.cs
```diff
@@ -35,5 +73,43 @@ public void HandleMessage(IServerMessageBase msg)
...
+        {
+            var memUsage = 0;
+            if (StoredMessagesData.Count > 0)
+            {
+                memUsage = StoredMessagesData.Count * StoredMessagesData.First().Value.GetMessageSize() / 1024; // This is in Kilobytes
```
DrewBanyai: This is an int, so dividing by 1024 means you'll get 0 until the value is at or above 1024, which is a lot of bytes to represent as 0 over the course of all the times you're going to write this debug message.

LmpClient/VesselUtilities/VesselCommon.cs
DrewBanyai: Small nitpick: UpdateTimestamp and LocalTimestamp should be cabelCase to match standards

LmpClient/Systems/VesselActionGroupSys/VesselActionGroupMessageHandler.cs
```diff
@@ -18,7 +24,38 @@ public void HandleMessage(IServerMessageBase msg)
...
                 return;
+            
+            // This code helps prevent time paradoxes.
+            // Why? Because this code makes sure that we only apply updates that aren't from the future.
+            // Additionally, to prevent us from getting desynchronised, we store updates from the future, so that we can apply them later.
+            var IsFromFuture = VesselCommon.UpdateIsFromFuture(msgData.GameTime, WarpSystem.Singleton.CurrentSubspaceTime);
 
+            if (IsFromFuture)
+            {
+
```
DrewBanyai: Small nitpick: blank lines left in

DrewBanyai (changes requested):
On top of my specific comments, one thing that really jumps out: 13 different copy-pastes of the HandleMessage / OnUpdate / LogQueuedMessagesSize stuff is a massive duplication. This should be extracted to a base class with the other classes derive from.

LmpClient/Systems/VesselPartSyncUiFieldSys/VesselPartSyncUiFieldMessageHandler.cs
BraveCaperCat2: At the same time, you were the one who suggested switching from `ConcurrentQueue` to `SortedList`:
Screenshot 2026-06-04 151605_modified
A "min-heap" would appear to be a binary tree and not something I can easily iterate through - I'm not even sure why you brought it up in the first place.

I'll add a `lock` for each `SortedList` to prevent the issues you mentioned. (`SortedList` doesn't appear to have a thread-safe variant)

LmpClient/Systems/VesselResourceSys/VesselResourceMessageHandler.cs
BraveCaperCat2: Yeah, I accidentally wrote `>` where I should have wrote `==`.
Also, btw, there are 14 handlers of that type, not 13.

LmpClient/Systems/VesselUpdateSys/VesselUpdate.cs
BraveCaperCat2: The `BodyName` field was required by some other part of the code - the project wouldn't build without it. The fact that it moved down a line is because both this branch and `LunaMultiplayer/master` added it back (was accidentally removed in a merge conflict) in different positions.

It can be safely ignored.

LmpClient/Systems/VesselActionGroupSys/VesselActionGroupMessageHandler.cs
BraveCaperCat2: Fixed.

LmpClient/VesselUtilities/VesselCommon.cs
BraveCaperCat2: > Small nitpick: UpdateTimestamp and LocalTimestamp should be cabelCase to match standards

*camelCase, not "cabelCase".
This issue has now been fixed.

LmpClient/Systems/VesselDecoupleSys/VesselDecoupleMessageHandler.cs
BraveCaperCat2: This is now fixed in all instances of this issue.

LmpClient/Systems/VesselResourceSys/VesselResourceMessageHandler.cs
BraveCaperCat2: Now fixed in all instances.

LmpClient/Systems/VesselPartSyncUiFieldSys/VesselPartSyncUiFieldMessageHandler.cs
BraveCaperCat2: This is now fixed.

LmpClient/VesselUtilities/VesselCommon.cs
BraveCaperCat2: Originally, I was going to use `10^-10` there, instead of `10^-15`, but I wanted to be extra sure that they're the same orbit.
I guess I was so sure that the function always returns false.
I've changed it to `10^-5` which should be better, as well as adding some error handling.

LmpClient/Systems/VesselPartSyncUiFieldSys/VesselPartSyncUiFieldMessageHandler.cs
BraveCaperCat2: I've done that for all instances in the latest commit.

BraveCaperCat2:
> On top of my specific comments, one thing that really jumps out: 13 different copy-pastes of the HandleMessage / OnUpdate / LogQueuedMessagesSize stuff is a massive duplication. This should be extracted to a base class with the other classes derive from.

I'm still not sure how to deal with that.
See my second comment on your first comment in your first review for more information.

BraveCaperCat2:
@DrewBanyai Currently waiting on a re-re-review
Next time, could you maybe only post a comment for each problem once, (if any) even if it's in multiple files? I don't really want to break github today

PR #670, opened 2026-05-08
option to freeze time whilst server is off

### LINKED ISSUE #543: Request: Provide an option to "pause" the universe on server stop
Background:

- OS: Windows 10
- KSP Version: 1.12.4
- LMP version (include the build number if you are using a nighly build): 0.29.0

---

Mods (remember that **mods are not supported**):  

N/A

---

While this isn't a "bug", I will write the expected behavior and current behavior from my perspective as a user.

Expected behaviour:  

Stopping the server should theoretically halt the universe; if I restart the server after 24 hours, the universe should be re-created in the same state I left it, from craft orbits to positions of celestial bodies.

Current behaviour:  

The server records certain times (`StartTime.txt`, `Subspace.txt`) to allow the universe to be simulated across time when the server is off. This means that if, for example (totally not a real experience) I were to have a craft on an intercept course with the Mun, and I stop the server and come back 24 hours later, the Mun has progressed in its orbit, leaving my craft nowhere near the Mun as I had left it.

There is a workaround that achieves my desired result (I think):
1. When stopping the server, record my machine's time (referred to as `lastStopTime`)
2. When starting the server again, record my machine's time again (referred to as `startTime`)
3. Calculate `timeDiff = startTime - lastStopTime`
4. Add `timeDiff` to the value stored in `StartTime.txt`

ShiralynDev:
Fixes #543 

Adding a warp option for the server to determine if it should jump ahead or stay at the time it was at shutdown.
Saving EndTime in a new Time.txt file that replaces StartTime.txt.

The TimeContext system is a black box to me and I was wondering if anyone knew a better way of implementing this. I just implemented the solution the same way the current workaround mentioned in #543 did.
I was also wondering if the old startTime file should be deleted upon creation of the new one.

### DIFF
--- a/Server/Settings/Definition/WarpSettingsDefinition.cs
+++ b/Server/Settings/Definition/WarpSettingsDefinition.cs
@@ -9,5 +9,7 @@ namespace Server.Settings.Definition
     {
         [XmlComment(Value = "Specify the warp Type. Values: None, Subspace")]
         public WarpMode WarpMode { get; set; } = WarpMode.Subspace;
+        [XmlComment(Value = "Tells the server to warp forward on startup to compensate for the time the server was down.")]
+        public bool WarpSyncOnStartup { get; set; } = true;
     }
 }
--- a/Server/Settings/Definition/WarpSettingsDefinition.cs
+++ b/Server/Settings/Definition/WarpSettingsDefinition.cs
@@ -9,7 +9,7 @@ namespace Server.Settings.Definition
     {
         [XmlComment(Value = "Specify the warp Type. Values: None, Subspace")]
         public WarpMode WarpMode { get; set; } = WarpMode.Subspace;
-        [XmlComment(Value = "Tells the server to warp forward on startup to compensate for the time the server was down.")]
-        public bool WarpSyncOnStartup { get; set; } = true;
+        [XmlComment(Value = "Tells the server to pause the time whilst off. If set to false the server will keep track of time and affect crafts")]
+        public bool PauseTimeWhileShutdown { get; set; } = false;
     }
 }
--- a/Server/System/TimeSystem.cs
+++ b/Server/System/TimeSystem.cs
@@ -54,6 +54,7 @@ namespace Server.System
                     LunaLog.Debug("Creating new time file, replacement of start time file");
                     TimeContext.StartTime = GetStoredStartTimeFromFile();
                     TimeContext.EndTime = LunaNetworkTime.UtcNow;
+                    File.Delete(OldStartTimeFile); // delete it since we will be using the new timefile in the future
                 }
                 else
                 {
--- a/Server/System/TimeSystem.cs
+++ b/Server/System/TimeSystem.cs
@@ -57,7 +57,7 @@ namespace Server.System
                 }
                 else
                 {
-                    LunaLog.Debug("Creating new start time file");
+                    LunaLog.Debug("Creating new time file");
                     TimeContext.StartTime = LunaNetworkTime.UtcNow;
                     TimeContext.EndTime = LunaNetworkTime.UtcNow;
                 }
--- a/Server/System/TimeSystem.cs
+++ b/Server/System/TimeSystem.cs
@@ -5,58 +5,112 @@ using System;
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using Server.Settings;
+using Server.Settings.Structures;
 
 namespace Server.System
 {
     public static class TimeSystem
     {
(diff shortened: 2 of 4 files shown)


ShiralynDev:
This now deletes the old startTime file upon loading it, avoiding two timefiles when the new system is switched over to.

DasSkelett:
A clear solution would probably be to store & sync the "elapsed time". However this would need some refactoring across server & client, and thus also break compatibility. Given these circumstances I agree that this might be the best solution.
One remark regarding the new setting's name & description: will that one be clear to users on what it does?
Or should we keep it a bit more plain, like `PauseTimeWhileShutdown` (default `false` in this case, or `ProgressTimeWhileShutdown` to keep the boolean logic the same way around), with comment "Progress game time even while server is not running" or similar?

ShiralynDev:
> A clear solution would probably be to store & sync the "elapsed time". However this would need some refactoring across server & client, and thus also break compatibility. Given these circumstances I agree that this might be the best solution. One remark regarding the new setting's name & description: will that one be clear to users on what it does? Or should we keep it a bit more plain, like `PauseTimeWhileShutdown` (default `false` in this case, or `ProgressTimeWhileShutdown` to keep the boolean logic the same way around), with comment "Progress game time even while server is not running" or similar?

Should be better now.

Server/Settings/Definition/WarpSettingsDefinition.cs
```diff
@@ -9,5 +9,7 @@ public class WarpSettingsDefinition
     {
         [XmlComment(Value = "Specify the warp Type. Values: None, Subspace")]
         public WarpMode WarpMode { get; set; } = WarpMode.Subspace;
+        [XmlComment(Value = "Tells the server to pause the time whilst off. If set to false the server will keep track of time and affect crafts")]
+        public bool PauseTimeWhileShutdown { get; set; } = false;
```
BraveCaperCat2, line 13:
```suggestion
        [XmlComment(Value = "If enabled, the server will jump ahead in time upon startup to what it would have been if the server didn't shut down.")]
        public bool JumpTimeOnStartup { get; set; } = true;
```
This should be changed to an enabled-by-default setting that, when turned on, will make the server account for offline time.

Server/System/TimeSystem.cs
```diff
@@ -5,59 +5,114 @@
...
-        private static string StartTimeFile { get; } = Path.Combine(ServerContext.UniverseDirectory, "StartTime.txt");
+        private static string TimeFile { get; } = Path.Combine(ServerContext.UniverseDirectory, "Time.txt");
+        private static string OldStartTimeFile { get; } = Path.Combine(ServerContext.UniverseDirectory, "StartTime.txt");
 
         public static void Reset()
         {
-            LoadSavedStartTime();
+            LoadSavedTimes();
+
+            if (WarpSettings.SettingsStore.PauseTimeWhileShutdown)
```
BraveCaperCat2, line 22:
```suggestion
            if (!WarpSettings.SettingsStore.JumpTimeOnStartup)
```
Account for the previously mentioned setting name change.
Also, btw, why do you have to change the start time to do this when you could just stop the time from being advanced based on the start time?

Server/Web/Structures/Settings/ServerWarpSettings.cs
```diff
@@ -5,5 +5,6 @@ namespace Server.Web.Structures.Settings
     public class ServerWarpSettings
     {
         public string WarpMode => WarpSettings.SettingsStore.WarpMode.ToString();
+        public string PauseTimeWhileShutdown => WarpSettings.SettingsStore.PauseTimeWhileShutdown.ToString();
```
BraveCaperCat2, line 8:
```suggestion
        public string JumpTimeOnStartup => WarpSettings.SettingsStore.JumpTimeOnStartup.ToString();
```
Account for the previously mentioned name change.

BraveCaperCat2 (changes requested):
I would like it if the new setting's meaning was inverted. Additionally, there may be a simpler way to write the logic for this change.

Server/Settings/Definition/WarpSettingsDefinition.cs
ShiralynDev: I think jump time is as unclear as my first setting that I was asked to change. The default is being kept as what it was before to avoid confusion, this could be changed but I would like @DasSkelett opinion on it aswell.

Server/System/TimeSystem.cs
ShiralynDev: I think it caused some problem and I just implemented the workarounds way of doing it. This would also cause problems if you wanted to enable it again since you would probably want to advance from now on and not from when the server was first started.

PR #676, opened 2026-05-11
Suggestions template

ShiralynDev:
Adding a suggestion template to allow for suggestions in the issues tab like most other repos do. This would make it more clear for new contributors what features the players want to see but also let developers make suggestions about where code could be improved or systems remade. This would also let users build on to suggestions or motivate them. The problem with using discord for suggestions is that they easily get lost, it's hard to expand on them and non discord contributors would not be able to see what is suggested. Suggestions and bugs would also be easy to tell apart due to their label and even quickly be sorted trough github's UI.

Feel free to discuss why this would be a good/bad idea. I'm all ears for motivations behind not having suggestions on github but I would like to hear some reasons that outweigh the benifits.

### DIFF
--- a/.github/ISSUE_TEMPLATE/SuggestionTemplate.md
+++ b/.github/ISSUE_TEMPLATE/SuggestionTemplate.md
@@ -1,0 +1,23 @@
+---
+name: Suggestion
+about: Create a new suggestion to tell us what could be improved.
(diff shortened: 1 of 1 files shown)


BraveCaperCat2 (approved):
Looks good.

PR #677, opened 2026-05-12
Add --data-directory and -d

OmegaMetor:
This adds the command line argument --data-directory [folder], with an alias of -d [folder] to the server.
These can be used to tell the server where to store runtime data, specifically, Config, logs, plugins, and Universe. When not specified, it will default to AppDomain.CurrentDomain.BaseDirectory. This default insures backwards compatibility.

Use of these options should make it easier for server owners to update, since it can clearly separate the state you'd want to keep from the server files themselves.

The argument parsing uses System.CommandLine, added as a dependency. This can be easily used in the future for more options if needed.

Thank you!

### DIFF
--- a/Server/Context/ServerContext.cs
+++ b/Server/Context/ServerContext.cs
@@ -25,10 +25,11 @@ namespace Server.Context
         public static bool UsePassword => !string.IsNullOrEmpty(GeneralSettings.SettingsStore.Password);
 
         public static Stopwatch ServerClock = new Stopwatch();
-        public static string UniverseDirectory = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Universe");
-        public static string ConfigDirectory = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Config");
-        public static string ModFilePath = Path.Combine(ConfigDirectory, "LMPModControl.xml");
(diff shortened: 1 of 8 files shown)


Server/Context/ServerContext.cs
ShiralynDev, line 1: I might be a bit harsh but I'm a fan of initializing variables on creation, even if its just ```public static DataDirectory = ""``` but preferably it would be init to the base directory.

Server/MainServer.cs
ShiralynDev, line 1: Could a help command be created as well? Could the error on missing argument be improved? Would be nice if instead of `Required argument missing for option: '-d'.` it showed the syntax.

ShiralynDev (changes requested):

ShiralynDev (approved):

ShiralynDev:
Everything looks good to me but I'm unsure about adding dependencies I think @gavazquez could answer that.

ShiralynDev:
I just noticed there is a build error with the github actions, could you take a look at it?

DasSkelett (approved):
Thanks, code looks good. Can you squash the commits a bit for merge? I believe 1, 2 and 4 make sense to be squashed, 3 can stay on its own.

OmegaMetor:
Should be all good now!

DasSkelett (approved):

PR #681, opened 2026-05-14
Clean up GeoIP2 geolocalization provider

DasSkelett:
This does some clean up and additional enhancements based on [Stronger retry for Ip location and added GeoIp2](https://github.com/LunaMultiplayer/LunaMultiplayer/commit/dbab939deea354bdd79b3b511fe9a2efffb81d56) for the master servers, by

- moving the new GeoIP2 provider out into a new `IGeolocalization` implementation
- renaming the environment variables to something LMP specific: `LMP_GEOIP2_LICENSE_KEY` and `LMP_GEOIP2_ACCOUNT_ID`
- making it handle missing env vars more gracefully
- making the `MaxMind.GeoIP2.WebServiceClient` persistent, creating it initially in a static constructor
- prioritizing the GeoIP2 lookup (if the user configured an API key etc, we also want to use it)
- adding a "cache lookup" from the `AddressCountries` dict as a fast path in `Server.Update()` so that for servers that re-appeared or share a public IP, we immediately set the country instead of scheduling it at the end of the queue where it might have to sit for a few hours

### DIFF
--- a/LmpMasterServer/Geolocalization/GeoIp2.cs
+++ b/LmpMasterServer/Geolocalization/GeoIp2.cs
@@ -1,0 +1,43 @@
+﻿using MaxMind.GeoIP2;
+using LmpMasterServer.Log;
+using System;
+using System.Net;
+using System.Threading.Tasks;
+
+namespace LmpMasterServer.Geolocalization
+{
+    internal class GeoIp2 : IGeolocalization
+    {
+
+        private static WebServiceClient client;
+
+        static GeoIp2()
+        {
+            var licenseKey = Environment.GetEnvironmentVariable("LMP_GEOIP2_LICENSE_KEY");
+            if (string.IsNullOrEmpty(licenseKey))
+                return;
+
+            if (!int.TryParse(Environment.GetEnvironmentVariable("LMP_GEOIP2_ACCOUNT_ID"), out var accountId))
+                return;
+
+            var clientOptions = (Microsoft.Extensions.Options.IOptions<WebServiceClientOptions>)new WebServiceClientOptions() { AccountId = accountId, LicenseKey = licenseKey };
+            client = new WebServiceClient(GeolocationHttpClient.GetClient(), clientOptions);
+            LunaLog.Debug($"GeoIP2 client setup successful, using account ID {accountId}");
+        }
+
+        public static async Task<string> GetCountryAsync(IPEndPoint externalEndpoint)
+        {
+            if (client == null)
+                return null;
+            try
+            {
+                var country = await client.CountryAsync(externalEndpoint.Address);
+                return country.Country.IsoCode;
+            }
+            catch (Exception)
+            {
+                return null;
+            }
+        }
+    }
+}
--- a/LmpMasterServer/Geolocalization/IpLocate.cs
+++ b/LmpMasterServer/Geolocalization/IpLocate.cs
@@ -1,5 +1,4 @@
-﻿using MaxMind.GeoIP2;
-using System;
+﻿using System;
 using System.Net;
 using System.Text.Json.Nodes;
 using System.Threading.Tasks;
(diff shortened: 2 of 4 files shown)


BraveCaperCat2:
The automatic build failed, so there's likely something wrong here.

DasSkelett:
> The automatic build failed, so there's likely something wrong here.

No that was just the AppVeyor build failing for PRs again due to the missing zip passwords. I've disabled the PR builds by now in the AppVeyor settings, but this won't clear the check here.

BraveCaperCat2 (approved):
Looks good.

ShiralynDev (approved):
LGTM

LmpMasterServer/Geolocalization/GeoIp2.cs
```diff
@@ -0,0 +1,43 @@
...
+        static GeoIp2()
+        {
+            var licenseKey = Environment.GetEnvironmentVariable("LMP_GEOIP2_LICENSE_KEY");
+            if (string.IsNullOrEmpty(licenseKey))
+                return;
+
+            if (!int.TryParse(Environment.GetEnvironmentVariable("LMP_GEOIP2_ACCOUNT_ID"), out var accountId))
+                return;
+
+            var clientOptions = (Microsoft.Extensions.Options.IOptions<WebServiceClientOptions>)new WebServiceClientOptions() { AccountId = accountId, LicenseKey = licenseKey };
```
DrewBanyai, line 23: Could you confirm you've tested this with credentials installed? As far as I can see, WebServiceClientOptions does not implement IOptions<> which isn't a problem with no credentials so this will compile but will likely fail if credentials are supplied.

LmpMasterServer/Geolocalization/GeoIp2.cs
```diff
@@ -0,0 +1,43 @@
...
+            client = new WebServiceClient(GeolocationHttpClient.GetClient(), clientOptions);
```
DrewBanyai, line 24: MaxMind has it's own HttpClient instance and we should probably be writing to it and sending it instead of using our local, which could (now or in the future) include details we don't want to be sharing with a third party.

LmpMasterServer/Geolocalization/GeoIp2.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace LmpMasterServer.Geolocalization
+{
+    internal class GeoIp2 : IGeolocalization
+    {
+
+        private static WebServiceClient client;
+
+        static GeoIp2()
+        {
+            var licenseKey = Environment.GetEnvironmentVariable("LMP_GEOIP2_LICENSE_KEY");
```
DrewBanyai, line 16: Given that we used to use just the variable name "license_key" we need to be sure that when this ships, we inform all Master Server owners to update their configuration or else it will run and they'll be unconfigured due to only the old variable being set.

LmpMasterServer/Geolocalization/GeoIp2.cs
```diff
@@ -0,0 +1,43 @@
...
+            if (string.IsNullOrEmpty(licenseKey))
+                return;
+
+            if (!int.TryParse(Environment.GetEnvironmentVariable("LMP_GEOIP2_ACCOUNT_ID"), out var accountId))
```
DrewBanyai, line 20: Given that we used to use just the variable name "account_id" we need to be sure that when this ships, we inform all Master Server owners to update their configuration or else it will run and they'll be unconfigured due to only the old variable being set.

LmpMasterServer/Geolocalization/GeoIp2.cs
```diff
@@ -0,0 +1,43 @@
...
+    internal class GeoIp2 : IGeolocalization
+    {
+
+        private static WebServiceClient client;
+
+        static GeoIp2()
+        {
+            var licenseKey = Environment.GetEnvironmentVariable("LMP_GEOIP2_LICENSE_KEY");
+            if (string.IsNullOrEmpty(licenseKey))
+                return;
```
DrewBanyai, line 18: We should be returning some sort of error string or error code instead of just returning early. Especially since the variable names are changing, we need to make sure the system itself informs the user if they are unconfigured instead of failing silently.

LmpMasterServer/Geolocalization/GeoIp2.cs
```diff
@@ -0,0 +1,43 @@
...
+
+            if (!int.TryParse(Environment.GetEnvironmentVariable("LMP_GEOIP2_ACCOUNT_ID"), out var accountId))
+                return;
```
DrewBanyai, line 21: We should be returning some sort of error string or error code instead of just returning early. Especially since the variable names are changing, we need to make sure the system itself informs the user if they are unconfigured instead of failing silently.

PR #683, opened 2026-05-16
Cleanup code formatting, add linting workflow using dotnet format

DasSkelett:
This PR cleans up the overall formatting of the code, and introduces a workflow to ensure it stays formatted as specified in the `.editorconfig`.

It is split into several commits:
1) Update the `.editorconfig` to insert trailing newlines at the end of `.cs` files (this is generally the standard and we already have this in many cases) and to trim trailing whitespace at the end of lines (especially otherwise empty lines)
2) Run `dotnet format` using above config, to ensure we are in a consistent, clean state. Exclude `Lidgren` and `uhttpsharp` projects as these are essentially vendored-in dependencies
3) Adjust `.gitattributes` to checkout files with CRLF line endings everywhere (including macOS & Linux systems), as this is what we also enforce in the `.editorconfig`
4) Run `git add --renormalize -u .` to normalize any files that currently have different line endings (i.e. LF)
5) Add a new workflow that checks if any files are not formatted correctly
6) Update dependencies in the workflow files while we're on it

To review, it's easier to only look at individual commit diffs at first, especially of 1, 3, 5, 6. 2 and 4 contain the big changes that clutter the diff.

---

Note that this will kind of break backporting commits from master to the stable release branch.
We can postponse merging the PR as a whole until we are about to release the 0.30, but maybe it makes sense to pull in the commits adjusting the editorconfig and gitattributes beforehand already, to ensure the incoming code is somewhat clean when people use editors/IDEs that auto-format based on them?


BraveCaperCat2:
Some of the files this PR changes still don't seem to have trailing newlines.

PR #687, opened 2026-05-16
Fix: docking local diffs need the same repair as remote docking diffs

AdmiralRadish:
This recreates the docking-local-diffs contribution as a clean single-commit PR.

It keeps the local undock guard/recovery behavior and includes the missing Linq import fix in Part_Undock so the branch compiles cleanly.

No unrelated files are included.


AdmiralRadish:
Tested on KerbalRealtime, undocking is working as intended between vessels with ports and docking ports pointed at fuel tanks etc. Merging

AdmiralRadish:
@Fierce-Cat FYI this may related to PR#660

PR #688, opened 2026-05-17
Avoid deprecated config in github action

ShiralynDev:
Switches from config to buildkitd-config as recommended.

Updates the workflow for docker-build.yml to avoid warning when running actions.

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -27,7 +27,7 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
         with:
-          config: .github/buildkitd.toml
+          buildkitd-config: .github/buildkitd.toml
       - name: Build container image
         uses: docker/build-push-action@v6
         with:


DasSkelett:
Found the same as part of #683 - https://github.com/LunaMultiplayer/LunaMultiplayer/commit/4813b6a164e992ba8a9d91923c2a83da438dd5fe
The same applies to docker-push.
Shall we extract that commit into a separate PR or this PR?

ShiralynDev:
> Found the same as part of #683 - [4813b6a](https://github.com/LunaMultiplayer/LunaMultiplayer/commit/4813b6a164e992ba8a9d91923c2a83da438dd5fe) The same applies to docker-push. Shall we extract that commit into a separate PR or this PR?

I'm fine with whatever you want. I can close this PR and we could just use yours if you want to. I did not notice you already made a PR solving this when I found the issue :P

BraveCaperCat2 (approved):
Looks good.

Issue #692, opened 2026-05-18
[BUG] XML Parser is case sensitive with booleans


BraveCaperCat2:
## Details:

This report was originally made on discord. None of the required details were explicitly provided, but the issue is pretty obvious.

Please provide your logs here:
```
[06:54:45][Fatal]: Error while trying to read /home/container/LMPServer/Config/GeneralSettings.xml. Default settings will 1 be used. Please remove the file so a new one can be generated.
```

And saves/configs here (if possible):
The general settings configuration file contained a `False` instead of `false`.

## Other mods:

Unsure

## Expected behaviour:

The user, @Splingold on discord, likely expected their changes in the `GeneralSettings.xml` file to have an effect, instead of the server throwing an error.

## Current behaviour:

The parser errors out due to the `False` not being considered a valid boolean.

## Steps to reproduce:

Replace one of the `true`s or `false`s with a `False`. (or potentially `True`)
Yeah, it's that simple.

ShiralynDev:
Half fixed with #693, closing the issue but if you are unhappy and still want full case insensitivity please reopen. I don't think we will be seeing bug reports due to "FaLsE" not working lol.

BraveCaperCat2:
> Half fixed with [#693](https://github.com/LunaMultiplayer/LunaMultiplayer/pull/693), closing the issue but if you are unhappy and still want full case insensitivity please reopen. I don't think we will be seeing bug reports due to "FaLsE" not working lol.

Yeah, this is completely fine.

PR #693, opened 2026-05-19
Allow common way of writing booleans in xml files

ShiralynDev:
Solves #692 atleast for "True" and "False" cases but not for "FaLsE" or "TrUe" cases.

Load the xml as text and replace ">True<" with ">true<" and same for false to allow this common way of writing booleans. This avoids confusion that users experience due to not knowing booleans should be lowercase. 

A regx could be used for even more control, for example allowing "FaLsE" but I did not see the need.

### DIFF
--- a/LmpCommon/Xml/LunaXmlSerializer.cs
+++ b/LmpCommon/Xml/LunaXmlSerializer.cs
@@ -56,7 +56,10 @@ namespace LmpCommon.Xml
                 return null;
             try
             {
-                using (TextReader r = new StreamReader(path))
+                var xml = File.ReadAllText(path);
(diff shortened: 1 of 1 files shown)


BraveCaperCat2 (approved):
This would appear to function as intended.

It would be nice to have it be truly case insensitive for booleans, but this should be enough for most if not all people.

ShiralynDev:
> This would appear to function as intended.
> 
> It would be nice to have it be truly case insensitive for booleans, but this should be enough for most if not all people.

I just hate writing regex :laughing: I will merge this but if someone wanna make it fully case insensitive they can do so in the future.

PR #694, opened 2026-05-19
Make 404 errors occur (when warranted) on master servers

BraveCaperCat2:
- Places where 404 errors should occur are now correctly handled
- Minor code cleanup for some related systems
- The master server now broadcasts the server list when the directory `serverlist` is provided and not just when no directory is provided

### DIFF
--- a/LmpMasterServer/Http/LunaHttpServer.cs
+++ b/LmpMasterServer/Http/LunaHttpServer.cs
@@ -26,14 +26,30 @@ namespace LmpMasterServer.Http
 
             FileHandler.HttpRootDirectory = WebHandler.BasePath;
 
+            // Exception handler, shows the user an error if an exception occurs
             Server.Use(new ExceptionHandler());
+
+            // Head handler, handles HTTP HEAD requests (like GET but without the content body)
             Server.Use(new HeadHandler());
+
+            // Compression handler
             Server.Use(new CompressionHandler(DeflateCompressor.Default, GZipCompressor.Default));
+
+            // Path guard, blocks path traversal attacks
             Server.Use(new PathGuard(WebHandler.BasePath));
+
+            // File handler, sends files if they're present on the server
             Server.Use(new FileHandler());
+
+            // HTTP router, selects between certain other handlers depending on the path
+            var listHandler = new ServerListHandler();
             Server.Use(new HttpRouter()
-                .With(string.Empty, new ServerListHandler())
-                .With("json", new RestHandler<ServerJson>(new ServerInfoRestHandler(), JsonResponseProvider.Default)));
+                .With(string.Empty, listHandler) // List handler, sends the server list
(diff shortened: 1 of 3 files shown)


LmpMasterServer/Http/LunaHttpServer.cs
```diff
@@ -26,14 +26,30 @@ public static void Start()
 
             FileHandler.HttpRootDirectory = WebHandler.BasePath;
 
+            // Exception handler, shows the user an error if an exception occurs
```
ShiralynDev, line 29: Are comments like this really needed. The class itself says the same thing basically. I'm a big fan of making code self explanatory in a way that avoids comments like these. I might be the odd one out here tho so please let me know your thoughts.

ShiralynDev (commented):
Looks good for the most part but I got some notes.

LmpMasterServer/Http/LunaHttpServer.cs
BraveCaperCat2: When I first looked at this area of the code, it wasn't very clear what each handler did and why they did it. Sure, the exception handler, file handler and server list handler (as well as the new `NotFoundHandler`) are pretty obvious and you can probably take an educated guess at the path guard being used to prevent path traversal attacks, (where a malicious user gains access to the file system by writing the URL to the server followed by a bunch of `../` in the address bar) but the head handler, compression handler and HTTP router are much less obvious. (I'm still not sure what the compression handler is for) I'm pretty sure that you need to actually view the source code to even attempt to understand what the HTTP router is doing.

BraveCaperCat2:
I force pushed the branch because when I pulled from `upstream/master`, my local git commits had different hashes from the ones on `origin/fix/lmpms-404`.

It also keeps the branch up-to-date with the latest copy of `upstream/master`.

LmpMasterServer/Http/LunaHttpServer.cs
ShiralynDev: I guess, I still think the comment does not add more detail. Comments like "// Compression handler" only repeat what the code says.

BraveCaperCat2: > I guess, I still think the comment does not add more detail. Comments like "// Compression handler" only repeat what the code says.

Yeah, because I'm still not sure exactly what it does and why.

DasSkelett: > > I guess, I still think the comment does not add more detail. Comments like "// Compression handler" only repeat what the code says.
> 
> Yeah, because I'm still not sure exactly what it does and why.

Most HTTP traffic is transferred compressed nowadays to save on bandwidth etc.
This handler runs the finished HTTP body through a compression function if support is advertised by the client (browser).

DasSkelett (approved):

PR #695, opened 2026-05-21
Fix the version checker

BraveCaperCat2:
- Fixes the version checker so that it won't log twice every 30 seconds whenever there's an update available.
- Changes the version checker to make a pair of log entries every minute
- Add a "multiplier" to multiply the time between pairs of log entries by 60 whenever the new version is considered compatible with the old version. (should be paired with a system to load in cross-compatible major version numbers directly from github)

### DIFF
--- a/Server/Utilities/VersionChecker.cs
+++ b/Server/Utilities/VersionChecker.cs
@@ -26,21 +26,23 @@ namespace Server.Utilities
         {
             while (ServerContext.ServerRunning)
             {
+                var multiplier = 1;
                 if (LatestVersion > LmpVersioning.CurrentVersion)
                 {
-                    LunaLog.Warning($"There is a new version of LMP! Please download it! Current: {LmpVersioning.CurrentVersion} Latest: {LatestVersion}");
+                    LunaLog.Info($"There is an update available for LMP, please download it when you're able to: {LmpVersioning.CurrentVersion} -> {LatestVersion}");
                     if (LmpVersioning.IsCompatible(LatestVersion))
                     {
-                        LunaLog.Debug("Your version is compatible with the latest version so you will still be listed in the master servers.");
+                        LunaLog.Info($"This update is not required to stay compattibile with updated master servers and clients.");
+                        multiplier = 60;
                     }
                     else
                     {
-                        LunaLog.Warning("Your version is NOT compatible with the latest version. You won't be listed in the master servers!");
+                        LunaLog.Warning("This update is mandatory to be shown on the server list once it's updated. You should go update");
                     }
                 }
 
-                //Sleep for 30 seconds...
-                await Task.Delay(30 * 1000);
+                // Repeat again in an hour if it's non-essential, or in a minute if it is essential.
+                await Task.Delay(60 * 1000 * multiplier);
             }
         }
     }
--- a/Server/Utilities/VersionChecker.cs
+++ b/Server/Utilities/VersionChecker.cs
@@ -37,7 +37,8 @@ namespace Server.Utilities
                     }
                     else
                     {
-                        LunaLog.Warning("This update is mandatory to be shown on the server list once it's updated. You should update the server ASAP.");
+                        LunaLog.Warning("This update is required in order to be shown on the server list and to connect with clients running the new version.\n"
+                        + "You should update the server ASAP.");
                     }
                 }
 
--- a/Server/Utilities/VersionChecker.cs
+++ b/Server/Utilities/VersionChecker.cs
@@ -26,17 +26,24 @@ namespace Server.Utilities
         {
             while (ServerContext.ServerRunning)
             {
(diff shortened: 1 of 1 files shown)


Server/Utilities/VersionChecker.cs
```diff
@@ -26,21 +26,23 @@ public static async Task DisplayNewVersionMsgAsync()
...
             {
+                var multiplier = 1;
                 if (LatestVersion > LmpVersioning.CurrentVersion)
                 {
-                    LunaLog.Warning($"There is a new version of LMP! Please download it! Current: {LmpVersioning.CurrentVersion} Latest: {LatestVersion}");
+                    LunaLog.Info($"There is an update available for LMP, please download it when you're able to: {LmpVersioning.CurrentVersion} -> {LatestVersion}");
                     if (LmpVersioning.IsCompatible(LatestVersion))
                     {
-                        LunaLog.Debug("Your version is compatible with the latest version so you will still be listed in the master servers.");
+                        LunaLog.Info($"This update is not required to stay compattibile with updated master servers and clients.");
```
DasSkelett: Typo in "compatible" ^^

Server/Utilities/VersionChecker.cs
```diff
@@ -26,21 +26,23 @@ public static async Task DisplayNewVersionMsgAsync()
...
                     if (LmpVersioning.IsCompatible(LatestVersion))
                     {
-                        LunaLog.Debug("Your version is compatible with the latest version so you will still be listed in the master servers.");
+                        LunaLog.Info($"This update is not required to stay compatible with updated master servers and clients.");
+                        multiplier = 60;
                     }
                     else
                     {
-                        LunaLog.Warning("Your version is NOT compatible with the latest version. You won't be listed in the master servers!");
+                        LunaLog.Warning("This update is mandatory to be shown on the server list once it's updated. You should update the server ASAP.");
```
ShiralynDev: This text was a bit hard to read for me. I think it could be restructured. I don't get the "once it's updated" part in the first sentence.

ShiralynDev (changes requested):

Server/Utilities/VersionChecker.cs
BraveCaperCat2: The "once it's updated" part is for when the release version of the master server is a release behind. (like the nightshade release-branch master server is right now, not having been updated to 0.29.2 yet)

To be fair, the "mandatory to be" part is sounding less gramatical to me now then when I wrote it, so I'll still change it.

Server/Utilities/VersionChecker.cs
BraveCaperCat2: Resolved.

ShiralynDev (approved):
LGTM

Server/Utilities/VersionChecker.cs
```diff
@@ -26,21 +26,24 @@ public static async Task DisplayNewVersionMsgAsync()
         {
             while (ServerContext.ServerRunning)
             {
+                var multiplier = 1;
                 if (LatestVersion > LmpVersioning.CurrentVersion)
                 {
-                    LunaLog.Warning($"There is a new version of LMP! Please download it! Current: {LmpVersioning.CurrentVersion} Latest: {LatestVersion}");
+                    LunaLog.Info($"There is an update available for LMP, please download it when you're able to: {LmpVersioning.CurrentVersion} -> {LatestVersion}");
                     if (LmpVersioning.IsCompatible(LatestVersion))
```
DrewBanyai, line 44: Since we're already tinkering with the magic number, I suggest we fix it a bit.

```
var delayMs = LmpVersioning.IsCompatible(LatestVersion)
    ? TimeSpan.FromHours(1)
    : TimeSpan.FromMinutes(1);
```

Then at the bottom:

```
await Task.Delay(delayMs);
```

DrewBanyai (approved):
Approved, though I'd recommend my one commented suggestion

Server/Utilities/VersionChecker.cs
BraveCaperCat2: Added in the latest commit.

DrewBanyai:
Appreciate the change but I have to nitpick a tiny bit more.

delayMs is a TimeSpan, so please rename it to just "delay".

For the ternary, please indent the two bottom lines as they are a continuation of the top line.

You don't need $ before a non-interpolating string.

Check if LatestVersion is null before the comparison, as it is null on startup and this could win the race condition and check null against a value with a > compare.

BraveCaperCat2:
> Appreciate the change but I have to nitpick a tiny bit more.
> 
> delayMs is a TimeSpan, so please rename it to just "delay".
> 
> For the ternary, please indent the two bottom lines as they are a continuation of the top line.
> 
> You don't need $ before a non-interpolating string.
> 
> Check if LatestVersion is null before the comparison, as it is null on startup and this could win the race condition and check null against a value with a > compare.

Fixed in the latest commit.

PR #697, opened 2026-05-27
Implement Science mode netcode and gameplay improvements

JogJogi:
Science sync bandwidth:
- Replace full ConfigNode binary (~800 bytes) with delta format (~50 bytes) for repeat experiment updates; full payload only sent on first discovery per subject
- Add WasTransmitted and CollectedBy metadata to every science subject message
- Add ScienceSubjectRevert message type: on revert-to-launch, client sends a pre-flight snapshot so server and other clients can roll back collected science; merge logic (science never decreases) protects concurrent players' progress

Science correctness:
- Server-side merge: incoming science value is ignored if it is lower than the current stored value, preventing race conditions when two players run the same experiment simultaneously
- Client-side deduplication via high-water-mark per subject; out-of-order or duplicate packets no longer cause unexpected science point changes
- Delta updates correctly apply to existing subjects; first-discovery full payloads are upserted regardless

Science Lab sync:
- Add dataProcessed field to ModuleScienceLab.xml so the lab processing state is included in the existing VesselPartSyncField sync path

Part UI field debounce:
- Buffer outgoing part UI field changes for 200ms before sending; rapid slider adjustments now coalesce into a single packet instead of flooding the network

Vessel position:
- Cache UpdateIntervalLockedToUnity and SecondaryVesselUpdatesUpdateIntervalLockedToUnity; values are recomputed only when Time.fixedDeltaTime changes instead of every frame
- Re-enable SendUnloadedSecondaryVesselPositionUpdates for abandoned/parked vessels

VesselProto stability:
- Wrap PrepareAndSendProtoVessel in try/catch; serialisation exceptions are logged with vessel id/name instead of silently crashing the background thread pool; non-debris vessels with 0-byte serialisation now produce a warning instead of being silently dropped

Group system GC:
- Remove intermediate List<string> allocations in AddMember, RemoveMember, and JoinGroup; build updated arrays directly and clone only once per operation

https://claude.ai/code/session_013Svqs8MSrPcfvLHosESs5F

### DIFF
--- a/LmpClient/ModuleStore/XML/Squad/ModuleScienceLab.xml
+++ b/LmpClient/ModuleStore/XML/Squad/ModuleScienceLab.xml
@@ -7,5 +7,8 @@
         <FieldDefinition>
             <FieldName>storedScience</FieldName>
         </FieldDefinition>
+        <FieldDefinition>
+            <FieldName>dataProcessed</FieldName>
+        </FieldDefinition>
     </Fields>
 </ModuleDefinition>
--- a/LmpClient/Network/NetworkReceiver.cs
+++ b/LmpClient/Network/NetworkReceiver.cs
@@ -276,6 +276,9 @@ namespace LmpClient.Network
                         case ShareProgressMessageType.ExperimentalPart:
                             ShareExperimentalPartsSystem.Singleton.EnqueueMessage(msg);
                             break;
+                        case ShareProgressMessageType.ScienceSubjectRevert:
+                            ShareScienceSubjectSystem.Singleton.EnqueueMessage(msg);
+                            break;
                     }
                     break;
                 case ServerMessageType.Screenshot:
--- a/LmpClient/Systems/Groups/GroupSystem.cs
+++ b/LmpClient/Systems/Groups/GroupSystem.cs
@@ -2,6 +2,7 @@
 using LmpClient.Network;
 using LmpClient.Systems.SettingsSys;
 using LmpCommon.Message.Data.Groups;
+using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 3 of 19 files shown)


BraveCaperCat2:
> Add ScienceSubjectRevert message type: on revert-to-launch, client sends a pre-flight snapshot so server and other clients can roll back collected science; merge logic (science never decreases) protects concurrent players' progress

This is a good idea, but the way it works relies on broken behaviour - see below.
> Server-side merge: incoming science value is ignored if it is lower than the current stored value, preventing race conditions when two players run the same experiment simultaneously

This makes it so that science never goes down, even when it's being spent on researching techs in the R&D facility.
> Buffer outgoing part UI field changes for 200ms before sending; rapid slider adjustments now coalesce into a single packet instead of flooding the network

200ms is a long time in programming. Maybe something like 50ms would be a better idea?
> Remove intermediate List allocations in AddMember, RemoveMember, and JoinGroup; build updated arrays directly and clone only once per operation

LunaMultiplayer does not have a group system yet. I don't know where you are getting this from, as it does not exist.
Additionally, @James on the discord server is already working on a companies system that would likely have the same affect as the "group system" that you've just imagined into existence. (and if not, I'm already planning on adding tons of new stuff into the companies system that would more than make up for it)
> ### Fixes included in this PR:
> ### Changes proposed in this PR:

You didn't utilise the PR template at all and didn't even clean it up.
> Screenshot 2026-05-27 193003_modified

You used AI to generate all of the commits in this PR, which doesn't make me very happy but is technically fine.

JogJogi:
Hey, sorry about that, i accidentally submitted this pull request while experimenting with git and didnt mean to open it. Thats why closed it as soon as i noticed. Apologies for that!

PR #698, opened 2026-05-31
Universe backup system

ShiralynDev:
Rename the current backup system to save.
Add a system for backups with an interval setting aswell as a setting for amount of backups to keep.
Saves backups as zipped folders in a Backups directory.

### DIFF
--- a/Server/Settings/Definition/GeneralSettingsDefinition.cs
+++ b/Server/Settings/Definition/GeneralSettingsDefinition.cs
@@ -81,7 +81,7 @@ namespace Server.Settings.Definition
         [XmlComment(Value = "Max number of parts that a vessel can have when spawning")]
         public int MaxVesselParts { get; set; } = 200;
 
-        [XmlComment(Value = "Max amount of backups to keep, minimum value of 1. BackupInterval in IntervalSettings can be used to turn off/on backups")]
+        [XmlComment(Value = "Max amount of backups to keep. 0 = Disabled. BackupInterval in IntervalSettings can be used to change how often to backup, it needs to be above 0 for backups to happen.")]
         public int MaxBackups { get; set; } = 5;
     }
 }
--- a/Server/Settings/Definition/GeneralSettingsDefinition.cs
+++ b/Server/Settings/Definition/GeneralSettingsDefinition.cs
@@ -81,7 +81,7 @@ namespace Server.Settings.Definition
         [XmlComment(Value = "Max number of parts that a vessel can have when spawning")]
         public int MaxVesselParts { get; set; } = 200;
 
-        [XmlComment(Value = "Max amount of backups to keep, minimum value of 0. BackupInterval in IntervalSettings can be used to turn off/on backups")]
+        [XmlComment(Value = "Max amount of backups to keep, minimum value of 1. BackupInterval in IntervalSettings can be used to turn off/on backups")]
         public int MaxBackups { get; set; } = 5;
     }
 }
--- a/Server/Settings/Definition/GeneralSettingsDefinition.cs
+++ b/Server/Settings/Definition/GeneralSettingsDefinition.cs
@@ -80,5 +80,8 @@ namespace Server.Settings.Definition
 
         [XmlComment(Value = "Max number of parts that a vessel can have when spawning")]
         public int MaxVesselParts { get; set; } = 200;
+
+        [XmlComment(Value = "Max amount of backups to keep, minimum value of 0. BackupInterval in IntervalSettings can be used to turn off/on backups")]
+        public int MaxBackups { get; set; } = 5;
     }
 }
--- a/Server/Settings/Definition/GeneralSettingsDefinition.cs
+++ b/Server/Settings/Definition/GeneralSettingsDefinition.cs
@@ -81,7 +81,7 @@ namespace Server.Settings.Definition
         [XmlComment(Value = "Max number of parts that a vessel can have when spawning")]
         public int MaxVesselParts { get; set; } = 200;
 
-        [XmlComment(Value = "Max amount of backups to keep. 0 = Disabled. BackupInterval in IntervalSettings can be used to change how often to backup, it needs to be above 0 for backups to happen.")]
+        [XmlComment(Value = "Max amount of backups to keep. 0 = Disabled. BackupMinutesInterval in IntervalSettings can be used to change how often to backup, it needs to be above 0 for backups to happen.")]
         public int MaxBackups { get; set; } = 5;
     }
 }
--- a/Server/Settings/Definition/IntervalSettingsDefinition.cs
+++ b/Server/Settings/Definition/IntervalSettingsDefinition.cs
@@ -24,7 +24,7 @@ namespace Server.Settings.Definition
         public int SaveIntervalMs { get; set; } = 30000;
 
         [XmlComment(Value = "Interval in ms at which internal LMP structures (Subspaces, Vessels, Scenario files, ...) will be backed up for longer storage. Set to 0 to avoid backups")]
-        public int BackupIntervalMs { get; set; } = 3600000;
+        public int BackupMinutesInterval { get; set; } = 60;
 
         [XmlComment(Value = "Interval to force a garbage collection and reduce the memory usage. Specify this value in minutes. 0 = deactivated.")]
         public int GcMinutesInterval { get; set; } = 15;
--- a/Server/Settings/Definition/IntervalSettingsDefinition.cs
+++ b/Server/Settings/Definition/IntervalSettingsDefinition.cs
@@ -23,6 +23,9 @@ namespace Server.Settings.Definition
         [XmlComment(Value = "Interval in ms at which internal LMP structures (Subspaces, Vessels, Scenario files, ...) will be saved to a file")]
         public int SaveIntervalMs { get; set; } = 30000;
 
+        [XmlComment(Value = "Interval in ms at which internal LMP structures (Subspaces, Vessels, Scenario files, ...) will be backed up for longer storage. Set to 0 to avoid backups")]
+        public int BackupIntervalMs { get; set; } = 0;
+
         [XmlComment(Value = "Interval to force a garbage collection and reduce the memory usage. Specify this value in minutes. 0 = deactivated.")]
         public int GcMinutesInterval { get; set; } = 15;
     }
--- a/Server/Settings/Definition/IntervalSettingsDefinition.cs
+++ b/Server/Settings/Definition/IntervalSettingsDefinition.cs
@@ -20,8 +20,8 @@ namespace Server.Settings.Definition
         [XmlComment(Value = "Main thread polling in ms. Keep this value low but at least above 2ms to avoid extreme CPU usage.")]
         public int MainTimeTick { get; set; } = 5;
 
-        [XmlComment(Value = "Interval in ms at which internal LMP structures (Subspaces, Vessels, Scenario files, ...) will be backed up to a file")]
-        public int BackupIntervalMs { get; set; } = 30000;
+        [XmlComment(Value = "Interval in ms at which internal LMP structures (Subspaces, Vessels, Scenario files, ...) will be saved to a file")]
+        public int SaveIntervalMs { get; set; } = 30000;
 
         [XmlComment(Value = "Interval to force a garbage collection and reduce the memory usage. Specify this value in minutes. 0 = deactivated.")]
         public int GcMinutesInterval { get; set; } = 15;
--- a/Server/Settings/Definition/IntervalSettingsDefinition.cs
+++ b/Server/Settings/Definition/IntervalSettingsDefinition.cs
@@ -24,7 +24,7 @@ namespace Server.Settings.Definition
         public int SaveIntervalMs { get; set; } = 30000;
 
         [XmlComment(Value = "Interval in ms at which internal LMP structures (Subspaces, Vessels, Scenario files, ...) will be backed up for longer storage. Set to 0 to avoid backups")]
-        public int BackupIntervalMs { get; set; } = 0;
+        public int BackupIntervalMs { get; set; } = 3600000;
 
         [XmlComment(Value = "Interval to force a garbage collection and reduce the memory usage. Specify this value in minutes. 0 = deactivated.")]
         public int GcMinutesInterval { get; set; } = 15;
--- a/Server/System/BackupSystem.cs
+++ b/Server/System/BackupSystem.cs
@@ -109,28 +109,38 @@ namespace Server.System
             }
         }
 
+        private static readonly object BackupLock = new();
+
         public static void RunBackup()
         {
             LunaLog.Debug("Performing backup...");
 
-            if (!Path.Exists(ServerContext.BackupDirectory))
+            if (!Directory.Exists(ServerContext.BackupDirectory))
                 Directory.CreateDirectory(ServerContext.BackupDirectory);
 
-            RemoveBackupsAboveLimit();
-            RunSave();
+            var tempUniversePath = Path.Combine(ServerContext.BackupDirectory, $"temp_backup{Guid.NewGuid():N}");
 
-            var tempUniversePath = Path.Combine(ServerContext.BackupDirectory, "temp_backup");
-            if (Directory.Exists(tempUniversePath))
-                Directory.Delete(tempUniversePath, recursive: true);
+            lock (BackupLock)
+            {
+                RemoveBackupsAboveLimit();
+                RunSave();
 
-            foreach (var dirPath in Directory.EnumerateDirectories(ServerContext.UniverseDirectory, "*", SearchOption.AllDirectories))
-                Directory.CreateDirectory(dirPath.Replace(ServerContext.UniverseDirectory, tempUniversePath));
+                foreach (var dirPath in Directory.EnumerateDirectories(ServerContext.UniverseDirectory, "*", SearchOption.AllDirectories))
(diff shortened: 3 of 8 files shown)


Server/Settings/Definition/IntervalSettingsDefinition.cs
```diff
@@ -20,8 +20,11 @@ public class IntervalSettingsDefinition
         [XmlComment(Value = "Main thread polling in ms. Keep this value low but at least above 2ms to avoid extreme CPU usage.")]
         public int MainTimeTick { get; set; } = 5;
 
-        [XmlComment(Value = "Interval in ms at which internal LMP structures (Subspaces, Vessels, Scenario files, ...) will be backed up to a file")]
-        public int BackupIntervalMs { get; set; } = 30000;
+        [XmlComment(Value = "Interval in ms at which internal LMP structures (Subspaces, Vessels, Scenario files, ...) will be saved to a file")]
+        public int SaveIntervalMs { get; set; } = 30000;
+
+        [XmlComment(Value = "Interval in ms at which internal LMP structures (Subspaces, Vessels, Scenario files, ...) will be backed up for longer storage. Set to 0 to avoid backups")]
+        public int BackupIntervalMs { get; set; } = 0;
```
BraveCaperCat2:
```suggestion
        public int BackupIntervalMs { get; set; } = 3600000;
```
Backups should be turned on by default. (here set to once every hour)
Also, why is the interval in milliseconds and not minutes? It feels like we're just introducing large numbers for the sake of it here.

Server/Settings/Definition/GeneralSettingsDefinition.cs
```diff
@@ -80,5 +80,8 @@ public class GeneralSettingsDefinition
 
         [XmlComment(Value = "Max number of parts that a vessel can have when spawning")]
         public int MaxVesselParts { get; set; } = 200;
+
+        [XmlComment(Value = "Max amount of backups to keep, minimum value of 0. BackupInterval in IntervalSettings can be used to turn off/on backups")]
```
BraveCaperCat2:
```suggestion
        [XmlComment(Value = "Max amount of backups to keep, minimum value of 1. BackupInterval in IntervalSettings can be used to turn off/on backups")]
```
It should say "minimum value of 1" instead, if BackupInterval is to be used in order to turn on/off backups. Personally, I disagree with using it like that, but feel free to do so.

Server/System/BackupSystem.cs
```diff
@@ -4,47 +4,133 @@
...
+using System.IO.Compression;
+using System.Security.Cryptography.X509Certificates;
 
 namespace Server.System
 {
     public class BackupSystem
     {
         //Subscribe to the exit event so a backup is performed when closing the server
-        static BackupSystem() => ExitEvent.ServerClosing += RunBackup;
+        static BackupSystem() => ExitEvent.ServerClosing += RunSave;
```
BraveCaperCat2, line 18: I'm not even sure what this is supposed to be doing.

Server/System/BackupSystem.cs
```diff
@@ -4,47 +4,133 @@
...
+                Directory.CreateDirectory(dirPath.Replace(ServerContext.UniverseDirectory, tempUniversePath));
+
+            foreach (var filePath in Directory.EnumerateFiles(ServerContext.UniverseDirectory, "*", SearchOption.AllDirectories))
+                File.Copy(filePath, filePath.Replace(ServerContext.UniverseDirectory, tempUniversePath), overwrite: true);
+
+            ZipFile.CreateFromDirectory(tempUniversePath, Path.Combine(ServerContext.BackupDirectory, $"backup_{DateTime.UtcNow:yyyy-MM-dd_HH-mm-ss}.zip"));
+            Directory.Delete(tempUniversePath, recursive: true);
+
+            LunaLog.Debug("Backups done");
+        }
```
BraveCaperCat2, line 152: This function is a mess from the perspective of the github splitscreen editor.

BraveCaperCat2 (changes requested):
I've noticed multiple improvements that should be made before this is pushed to `master`.

Server/Settings/Definition/IntervalSettingsDefinition.cs
ShiralynDev: Could you not argue the same for save interval? I'm just trying to keep it consistent.

Server/System/BackupSystem.cs
ShiralynDev: this was there before, it saves the game when the server is closed.

Server/Settings/Definition/IntervalSettingsDefinition.cs
BraveCaperCat2: > Could you not argue the same for save interval? I'm just trying to keep it consistent.

The save interval, however, is only 30 seconds...

BraveCaperCat2 (approved):
Looks good.
I still want to argue that the backup interval setting should be in minutes, but that's only a minor issue.

Server/Settings/Definition/IntervalSettingsDefinition.cs
DasSkelett: I also agree with using a useful unit like minutes, I don't think the sake of consistency is worth it here.
It'll only cause people to set unrealistic values here, and we end up trying to do back ups every 30ms... so unless we change it to minutes, we probably also need a larger minimum value and enforce it correspondingly.

ShiralynDev: > I also agree with using a useful unit like minutes, I don't think the sake of consistency is worth it here. It'll only cause people to set unrealistic values here, and we end up trying to do back ups every 30ms... so unless we change it to minutes, we probably also need a larger minimum value and enforce it correspondingly.

Will change then.

ShiralynDev: > > Could you not argue the same for save interval? I'm just trying to keep it consistent.
> 
> The save interval, however, is only 30 seconds...

Yea, I did not notice that at first, thought it was around 5 min or something.

Server/Settings/Definition/IntervalSettingsDefinition.cs
```diff
@@ -24,7 +24,7 @@ public class IntervalSettingsDefinition
         public int SaveIntervalMs { get; set; } = 30000;
 
         [XmlComment(Value = "Interval in ms at which internal LMP structures (Subspaces, Vessels, Scenario files, ...) will be backed up for longer storage. Set to 0 to avoid backups")]
-        public int BackupIntervalMs { get; set; } = 3600000;
+        public int BackupMinutesInterval { get; set; } = 60;
```
BraveCaperCat2, line 27: I would call it `BackupIntervalMin` but this looks fine.

Server/System/BackupSystem.cs
```diff
@@ -4,47 +4,130 @@
...
                 }
-                catch (TaskCanceledException)
+                catch (TaskCanceledException) { break; }
+            }
+        }
+
+        private static async Task RunBackupsAsync(CancellationToken token)
+        {
+            if (IntervalSettings.SettingsStore.BackupMinutesInterval == 0 || GeneralSettings.SettingsStore.MaxBackups == 0) {
+                LunaLog.Normal("Backups are off");
```
BraveCaperCat2, line 48:
```suggestion
                LunaLog.Warning("Backups are turned off. You can turn them on again in the server configuration.");
```
This should probably be a warning, rather than a normal log entry.

BraveCaperCat2 (changes requested):

Server/Settings/Definition/IntervalSettingsDefinition.cs
ShiralynDev: Min can be read as minimum and see to line below.

Server/System/BackupSystem.cs
ShiralynDev: Why? Not using backups is a normal choice.

Server/System/BackupSystem.cs
```diff
@@ -4,47 +4,130 @@
...
+                            CultureInfo.InvariantCulture, DateTimeStyles.None, out var dt))
+                        return new { f.path, Date = (DateTime?)dt };
+
+                    return new { f.path, Date = (DateTime?)null };
+                })
+                .Where(f => f.Date.HasValue)
+                .OrderByDescending(f => f.Date)
+                .ToList();
+
+            while (backups.Count >= GeneralSettings.SettingsStore.MaxBackups)
```
DrewBanyai, line 110: If MaxBackups is 0, this while() will always resolve to true, but then two lines down you reference backups.Last which will return an InvalidOperationException since it will be an empty list...

A simple solution would be to return out early if MaxBackups is 0

In addition, create a test plan since this error indicates the changes weren't thoroughly tested.

DrewBanyai (changes requested):

Server/System/BackupSystem.cs
```diff
@@ -4,47 +4,130 @@
...
+        {
+            LunaLog.Debug("Performing backup...");
+
+            if (!Path.Exists(ServerContext.BackupDirectory))
+                Directory.CreateDirectory(ServerContext.BackupDirectory);
+
+            RemoveBackupsAboveLimit();
+            RunSave();
+
+            var tempUniversePath = Path.Combine(ServerContext.BackupDirectory, "temp_backup");
```
DrewBanyai: If you look at RunSave, you can see it takes the LockObj to ensure saves are consistent and do not run into race conditions. The below code happens outside of that LockObj and also runs asynchronously on a scheduled RunSavesAsync which means it's a race condition waiting to happen. Run the server with this change for a few hours and you're likely to hit RunBackup at a time when the LockObj is being pulled from another thread and the entire universe backup will produce an inconsistent and likely corrupt backup, probably throwing an IOException to boot.

The snapshot should be copied under the LockObj to avoid this.

Server/System/BackupSystem.cs
```diff
@@ -4,47 +4,130 @@
...
+            if (Directory.Exists(tempUniversePath))
```
DrewBanyai: Having temp_backup be a constant path means that a scheduled save happening at the same time as a manual /backup save will stomp each other (a Directory.Delete happening while another one is copying). Either use a unique temp directory name, or serialize the entire RunBackup to avoid this.

Server/System/BackupSystem.cs
```diff
@@ -4,47 +4,130 @@
...
+            while (ServerContext.ServerRunning)
+            {
+                try
                 {
-                    break;
+                    await Task.Delay(IntervalSettings.SettingsStore.BackupMinutesInterval*60000, token);
                 }
+                catch (TaskCanceledException) { break; }
+
+                RunBackup();
```
DrewBanyai: This should also be within a try/catch since there are numerous ways this can fail.

Server/System/BackupSystem.cs
```diff
@@ -4,47 +4,130 @@
...
+                LunaLog.Normal("Backups are off");
+                return;
+            }
+
+            while (ServerContext.ServerRunning)
+            {
+                try
                 {
-                    break;
+                    await Task.Delay(IntervalSettings.SettingsStore.BackupMinutesInterval*60000, token);
```
DrewBanyai: This overflows int above ~35,791 minutes, and the codebase already has an established pattern in GcSystem

```await Task.Delay((int)TimeSpan.FromMinutes(IntervalSettings.SettingsStore.GcMinutesInterval).TotalMilliseconds, token);```

Please match it (and add spaces around the operator)

File: Server/System/BackupSystem.cs [hunk 12]
```diff
@@ -4,47 +4,130 @@
 using Server.Settings.Structures;
 using System.Threading;
 using System.Threading.Tasks;
+using System.IO;
+using System.Linq;
+using System;
+using System.Globalization;
+using System.IO.Compression;
+using System.Security.Cryptography.X509Certificates;
```
DrewBanyai: Why was X509Certificates included? Was this an accidental addition?

Server/System/BackupSystem.cs
```diff
@@ -4,47 +4,130 @@
 using Server.Settings.Structures;
 using System.Threading;
```
DrewBanyai, line 5: .Threading might no longer be needed based on the changes. Please run a build without it to confirm.

Server/Settings/Definition/IntervalSettingsDefinition.cs
```diff
@@ -20,8 +20,11 @@ public class IntervalSettingsDefinition
         [XmlComment(Value = "Main thread polling in ms. Keep this value low but at least above 2ms to avoid extreme CPU usage.")]
         public int MainTimeTick { get; set; } = 5;
 
-        [XmlComment(Value = "Interval in ms at which internal LMP structures (Subspaces, Vessels, Scenario files, ...) will be backed up to a file")]
-        public int BackupIntervalMs { get; set; } = 30000;
+        [XmlComment(Value = "Interval in ms at which internal LMP structures (Subspaces, Vessels, Scenario files, ...) will be saved to a file")]
+        public int SaveIntervalMs { get; set; } = 30000;
```
DrewBanyai, line 24: Every server that currently has BackupIntervalMs set in their IntervalSettingsDefinition will fall back to the default (30000) if this goes out and they'll have to update their settings, likely with no warning.

This is an unacceptable trade-off. Either use the existing variable name or retain the existing variable for a set amount of time and use it as a fallback if the new one isn't set so that server operators have time to transition instead of their automated backups all turning off because of a release.

Server/Settings/Definition/GeneralSettingsDefinition.cs
```diff
@@ -80,5 +80,8 @@ public class GeneralSettingsDefinition
 
         [XmlComment(Value = "Max number of parts that a vessel can have when spawning")]
         public int MaxVesselParts { get; set; } = 200;
+
+        [XmlComment(Value = "Max amount of backups to keep. 0 = Disabled. BackupInterval in IntervalSettings can be used to change how often to backup, it needs to be above 0 for backups to happen.")]
```
DrewBanyai: This comment has the wrong variable name "BackupInterval"

DrewBanyai (changes requested):

Server/System/BackupSystem.cs
BraveCaperCat2: In future, you should note that code suggestions can be added to comments.
```suggestion
                    await Task.Delay((int)TimeSpan.FromMinutes(IntervalSettings.SettingsStore.GcMinutesInterval).TotalMilliseconds, token);
```

Server/Settings/Definition/IntervalSettingsDefinition.cs
BraveCaperCat2: It should be noted that this would be in the patch notes for whatever update is getting these changes, so there would in fact be some warning.

Server/Settings/Definition/GeneralSettingsDefinition.cs
BraveCaperCat2:
```suggestion
        [XmlComment(Value = "Max amount of backups to keep. 0 = Disabled. BackupMinutesInterval in IntervalSettings can be used to change how often to backup, it needs to be above 0 for backups to happen.")]
```

Server/System/BackupSystem.cs
ShiralynDev: This would not be called tho as if backups are set to 0 it would not preform any backups?

ShiralynDev: There is a command for making backups, forgot about that...

Server/System/BackupSystem.cs
ShiralynDev: LunaMultiplayer/Server/System/BackupSystem.cs(44,51): error CS0246: The type or namespace name 'CancellationToken' could not be found (are you missing a using directive or an assembly reference?)

Server/Settings/Definition/IntervalSettingsDefinition.cs
ShiralynDev: I agree with @BraveCaperCat2 here. Calling the save "BackupIntervalMs" was misleading to begin with but more so with a real backup system. It could be kept but that would lead to BackupIntervalMs and Minutes coexisting or having a stupid name for the real backup system. A transition would have the same problem, people who don't read the important notes for major release changelogs would not notice and change their setting during a transition peroid either, the hit would just happen later.

Server/System/BackupSystem.cs
ShiralynDev: f973afd

Server/System/BackupSystem.cs
ShiralynDev: 19385f4

Server/System/BackupSystem.cs
ShiralynDev: ccadcf00eeb63644dc5d1cbb06c7c4734268804e

Server/System/BackupSystem.cs
ShiralynDev: Hope 205a34154b6b64874542a92c58f9110b45d917d0 is a good solution to you.

Server/System/BackupSystem.cs
ShiralynDev: Is the BackupLock in 205a34154b6b64874542a92c58f9110b45d917d0 enough?

Server/System/BackupSystem.cs
ShiralynDev: Fixed with a warn and return ff662b10087361cc650cf285f50bca04696cd773 this still allows you to run backups since you probably want that if you are running the command.

PR #699, opened 2026-06-01
Fix incorrect version numbers in 11 files

BraveCaperCat2:
- Fix the nightly builds being labelled as `0.29.3` when they're actually `0.30.0`
- Fix the issue template having a `0.29.1.x` version number being replaced with `0.29.3.x` (a version that doesn't actually exist)
- Changes made for the above fixes
- Nothing else


DrewBanyai (approved):
lgtm

ShiralynDev (approved):
Looks good

PR #702, opened 2026-06-05
Ignore non-local addresses

### LINKED ISSUE #653: ["BUG"]
## Details: Friend is able to connect to my server from across the globe. But im not able to connect to my own localy hosted pc. be it trough the interweb or with local ip. "Disconnected: Failed to establish connection - no response from remote host"

Which OS are you using? (e.g. Windows, Mac, Linux):Tried on both Windows and Linux
What version of KSP are you running? (e.g. 1.12.5): 1.12.5.31990
What version of LMP are you playing? (e.g. 0.29.0, 0.29.1.757): 0.29.0
Which KSP DLCs do you have? (e.g. Breaking Ground, Making History): Breaking ground and making history

Please provide your logs here:

[KSP.log](https://github.com/user-attachments/files/26941555/KSP.log)

Didnt find output_log.txt in appdata so here is this:
[Player.log](https://github.com/user-attachments/files/26941556/Player.log)

And saves/configs here (if possible):
Save was a new game and config only changed the name of the server and discription. also changed from sandbox to carrer

## Other mods:

no mods

## Expected behaviour:

Join the server

## Current behaviour:

not join the server

## Steps to reproduce:

start server on one pc in the local network and then try to join it with an other pc in the same network

BraveCaperCat2:
- Fix #653
- The master server now checks whether the internal endpoint is a local IP address
- If not, it sets the internal IPv4 and IPv6 addresses to `255.255.255.255` (and the IPv6 equivalent)
- The client then checks if the appropriate addresses are `255.255.255.255` (or the IPv6 equivalent) or not and only connects on LAN if it isn't.

### DIFF
--- a/LmpClient/Network/NetworkServerList.cs
+++ b/LmpClient/Network/NetworkServerList.cs
@@ -109,7 +109,10 @@ namespace LmpClient.Network
                 var amListeningOnIPv6 =
                     NetworkMain.ClientConnection.Socket.AddressFamily == AddressFamily.InterNetworkV6;
 
-                if (ServerIsInLocalLan(serverInfo.ExternalEndpoint) || (amListeningOnIPv6 && ServerIsInLocalLan(serverInfo.InternalEndpoint6)))
+                // Ignore the 255.255.255.255 address (and IPv6 equivalent) as a local address
+                bool isValidInternalEndpoint = !(serverInfo.InternalEndpoint.Address.Equals(IPAddress.None) || (amListeningOnIPv6 && serverInfo.InternalEndpoint6.Address.Equals(IPAddress.IPv6None)));
+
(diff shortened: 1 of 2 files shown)


DasSkelett (changes requested):
Please see https://github.com/LunaMultiplayer/LunaMultiplayer/issues/653#issuecomment-4629973900

Also we are currently using the loopback address as "placeholder" for "None" in most places, we should probably stick to that.

BraveCaperCat2:
> Please see [#653 (comment)](https://github.com/LunaMultiplayer/LunaMultiplayer/issues/653#issuecomment-4629973900)
> 
> Also we are currently using the loopback address as "placeholder" for "None" in most places, we should probably stick to that.

Trying to specifically find a private local address on the server seems quite difficult and potentially unnecessary.
Do we really need that?

I'll change `IPAddress.None` (and `IPAddress.IPv6None`) to `IPAddress.Loopback` (and `IPAddress.IPv6Loopback`) then

PR #706, opened 2026-06-19
Merge 0.29.2 in to master

ShiralynDev:
Merge 0.29.2 in to master to be able to develop on master and branch releases of it. This fixes our current versioning so release can be made more frequently and in a uniform way.


BraveCaperCat2:
@ShiralynDev The head branch for this should probably be a new branch so that we don't change the actual `master` branch when fixing merge conflicts. (that can then be fast-forwarded into `master` when we're ready)

ShiralynDev:
> @ShiralynDev The head branch for this should probably be a new branch so that we don't change the actual `master` branch when fixing merge conflicts. (that can then be fast-forwarded into `master` when we're ready)

That is true, will fix.

ShiralynDev:
See #707

PR #708, opened 2026-06-20
0_29_3-PR1: Core Runtime and Network Fixes

AdmiralRadish:
Cherry-picked/layered PR1 content from AdmiralRadish fork for 0_29_3 backport series
Scope:
- NetConnection NatPunch handling
- VesselPartSyncFieldMsgData deserialization fix
- Banned IP retriever/entrypoint runtime fixes

PR #709, opened 2026-06-20
0_29_3-PR2: Vessel Runtime Stability

AdmiralRadish:
Cherry-picked/layered PR2 content from AdmiralRadish fork for 0_29_3 backport series.
Scope:

- Vessel load hardening and load-outcome handling (fresh load, reload, unchanged, failure paths)
- Topology race protections around couple/decouple/undock mutation windows
- Proto drain/backpressure improvements to reduce frame spikes during burst reloads
- Scene-aware proto application behavior (full reload vs lightweight proto-swap)
- Drift broadcast gating to avoid repeated identical vessel-definition churn
- Vessel sync diagnostics/tracing for arrival, discard, defer, swap, load/reload outcomes
- Tracking station profiling instrumentation to identify scene-entry rebuild storms
- Client Harmony safety patches for date/crew/part registration and orbit update edge cases
- Discovery/serialization sanitization and vessel-loader robustness improvements
- ShareFunds correctness fixes for rollout-cost tracking, revert refund consistency, and scene reseeding
- Maneuver-node and vessel-definition sync follow-up coverage in vessel proto events
- Settings/plumbing updates to enable and control new diagnostics
- Comment cleanup pass to keep explanations concise while preserving intent


BraveCaperCat2:
Issue in formatting, see my reply to #708. Corrected description:
> Cherry-picked/layered PR2 content from AdmiralRadish fork for 0_29_3 backport series.
> 
> Scope:
> - Vessel load hardening and outcomes
> - Topology drift/race protections
> - Client harmony safety patches and diagnostics

LmpClient/Systems/KscScene/KscSceneSystem.cs
BraveCaperCat2, line 1: The comments in this file could probably be a bit shorter.

LmpClient/Systems/ShareFunds/ShareFundsEvents.cs
BraveCaperCat2, line 1: Could have shorter comments, but otherwise looks fine.

LmpClient/Systems/VesselProtoSys/VesselProtoEvents.cs
BraveCaperCat2, line 1: Comments could be shorter.

LmpClient/Systems/VesselProtoSys/VesselProtoSystem.cs
BraveCaperCat2, line 1: There are some really long comments in this file that should probably be made more consice.

BraveCaperCat2 (changes requested):
There are many comments that could (and probably should) be more concise.
The code and comments appear to do a lot more than what's listed in the PR description, including but likely not limited to:
- Tracking station profiling (I wasn't made aware of any ongoing performance issues with the tracking station?)
- Debug information for Funds-related code

I haven't viewed any of the larger diffs (e.g. file creation) in order to accelerate my review, I may need to do a more complete one at a later date.

AdmiralRadish:
>     Tracking station profiling (I wasn't made aware of any ongoing performance issues with the tracking station?)
>     Debug information for Funds-related code

I did a final commit to remove tracking-station profiling instrumentation (part of an issue with vessel drift reported in chat) and remove the funds-change log line, while keeping the actual stabilization logic.

BraveCaperCat2 (approved):
Thanks for making the comments more concise. Although the long comments provided more information, it can make the code a bit harder to read since less of it is actual code.

LmpClient/MainSystem.cs
```diff
@@ -319,8 +387,8 @@ public void HandleException(Exception e, string eventName)
         public void DisconnectFromGame()
         {
             ForceQuit = true;
-            ScenarioSystem.Singleton.SendScenarioModules();
             NetworkConnection.Disconnect("Quit");
+            ScenarioSystem.Singleton.SendScenarioModules();
```
DrewBanyai, line 391: I may be wrong, but it seems like this is going to break the saving of the scenario in most cases. Disconnect() is going to tear down the connection and then SendScenarioModules() is going to attempt to send the data to the server and fail.

I'm pretty sure the Disconnect calls ResetConnectionStaticsAndQueues(), right?

This was likely a regression in 0.29.2 which should be reversed. I apologize, I should have caught it in my branch.

LmpClient/Systems/VesselProtoSys/VesselProtoSystem.cs
```diff
@@ -64,9 +65,6 @@ protected override void OnEnabled()
 
             WarpEvent.onTimeWarpStopped.Add(VesselProtoEvents.WarpStopped);
 
-            GameEvents.onManeuverAdded.Add(VesselProtoEvents.ManeuverNodeAdded);
```
DrewBanyai, line 67: Could you explain this one just so I make sure I'm understanding it? It looks like this removes the callback for onManeuverAdded and onManeuverRemoved so when maneuver nodes change, no update is sent out. That means the updates only go out during full vessel syncs. It's very likely this could be intentional but I just want to make sure before I approve.

LmpClient/MainSystem.cs
```diff
@@ -77,13 +77,70 @@ public static ClientState NetworkState
...
+        ///         for exactly that long.</item>
+        ///   <item>The frame-count delta between two lines, divided by the
+        ///         timestamp delta, gives a no-tool fps estimate.</item>
+        ///   <item><c>dt</c> is the previous frame's
+        ///         <see cref="Time.unscaledDeltaTime"/>; a value much above
+        ///         ~0.5s in steady-state means one very expensive frame just
+        ///         landed (scene-load tail, vessel storm, etc.).</item>
+        /// </list>
+        /// </summary>
+        private static void EmitHeartbeatIfDue()
```
DrewBanyai, line 108: I know this is going to be annoying for me to say since I'm the one that put the Heartbeat debug log in, but this was never actually intended to reach master branch. It was for 0.29.2 only to help detect some of the issues that were causing slow-downs.

We absolutely can include it in master if it's wanted there and in future builds thereafter, but it should be a conscious decision so I'm just letting you know: Inclusion of this feature is at your discretion. I'll be fine if it's removed, since it served it's original purpose, is fine to stay in 0.29.2, and causes quite a bit of extra logging so some people may want it removed.

Up to you, just letting you know.

LmpClient/MainSystem.cs
BraveCaperCat2: This change is also on the 0.29.2 branch, see here:
https://github.com/LunaMultiplayer/LunaMultiplayer/blob/5f030f6762567bc82a9a3dd556a6490f0e5b20b1/LmpClient/MainSystem.cs#L397-L399

LmpClient/Systems/VesselProtoSys/VesselProtoSystem.cs
BraveCaperCat2: This change is also on the 0.29.2 branch, see here:
https://github.com/LunaMultiplayer/LunaMultiplayer/blob/5f030f6762567bc82a9a3dd556a6490f0e5b20b1/LmpClient/Systems/VesselProtoSys/VesselProtoSystem.cs#L95-L98

LmpClient/MainSystem.cs
BraveCaperCat2: As you already pointed out, this change is also on the 0.29.2 branch:
https://github.com/LunaMultiplayer/LunaMultiplayer/blob/5f030f6762567bc82a9a3dd556a6490f0e5b20b1/LmpClient/MainSystem.cs#L92-L134
Perhaps it could be put into debug build releases only from now on?

LmpClient/Systems/VesselProtoSys/VesselProtoSystem.cs
DrewBanyai: onManeuverAdded and onManeuverRemoved have actually never been in 0.29.2 so I think this might actually be something that was added in master and this PR is removing it, which may be a mistake since it was added separately into master from someone else's commit...

DrewBanyai: Comes from this change, by Radish himself: https://github.com/LunaMultiplayer/LunaMultiplayer/commit/5f899de084bae6ab4b99e8ebd36ca33405dd8976

LmpClient/Systems/ShareFunds/ShareFundsEvents.cs
```diff
@@ -7,9 +7,19 @@ public class ShareFundsEvents : SubSystem<ShareFundsSystem>
...
+            {
+                var delta = System.LastKnownFunds.Value - funds;
+                if (reason == TransactionReasons.VesselRollout && delta > 0)
+                {
+                    System.CurrentShipCost = new Tuple<Guid, float>(Guid.Empty, (float)delta);
+                }
+            }
+            System.LastKnownFunds = funds;
 
-            LunaLog.Log($"Funds changed to: {funds} reason: {reason}");
```
ShiralynDev, line 12: Why is the lunalog removed here?

ShiralynDev (approved):
LGTM

DrewBanyai (changes requested):
Stacking the remaining PRs will leave 2 errors in the build, and one is in this PR so I'm going to request just that one change before approving.

LmpClient/MainSystem.cs  (line 362) calls VesselSyncDiagnostics.Flush() which lives in LmpClient.Diagnostics so you need to add this at the top of MainSystem.cs:

```using LmpClient.Diagnostics```

DrewBanyai (approved):
LGTM :)

PR #716, opened 2026-07-07
Fix duplicate research node purchase by closing the R&D panels on other clients  (#667)

### LINKED ISSUE #667: ["BUG"] Possible to research same node twice
## Details:

Which OS are you using? (e.g. Windows, Mac, Linux): Linux (Kubuntu 25.10)
What version of KSP are you running? (e.g. 1.12.5): 1.12.5
What version of LMP are you playing? (e.g. 0.29.0, 0.29.1.757): Self-compiled on latest main branch (commit 558b5bc)
Which KSP DLCs do you have? (e.g. Breaking Ground, Making History): Breaking Ground, Making History

Please provide your logs here:
[KSP.log](https://github.com/user-attachments/files/27495139/KSP.log)

And saves/configs here (if possible):
[Universe.zip](https://github.com/user-attachments/files/27495233/Universe.zip)
[Config.zip](https://github.com/user-attachments/files/27495236/Config.zip)

## Expected behaviour:

Sync the nodes to show all players that the node has been purchased by someone.

## Current behaviour:

It syncs the main tree showing its unlocked but does not update the node details on the right side if it's already opened. It is also possible to buy twice or probably however many players you are since it can be purchased as the games thinks it isn't (on the details part atleast) and it will charge the science cost.

## Steps to reproduce:

Connect multiple clients to the server
Enter R&D with at least 2 clients
Open a specific research node on multiple clients
Buy the node one client
Buy it again on another

GrahamKowalski:
Fixes #667 

## Problem
When one player researches a tech node while other players have that same node open/selected in the R&D screen, the other clients still show the purchase button. Clicking it purchased the node again and spends science a second time.

## Cause
The client applies the research with `ResearchAndDevelopment.UnlockProtoTechNode()` and calls the static `ResearchAndDevelopment.RefreshTechTreeUI()`. That respawns the tech-tree nodes but didnot refresh open node panels in other clients, and strands `RDController.Instance.node_selected` 

## Fix
When the node being purchased is selected by other clients, close the node panel `ShowNothingPanel()` so the stale Research button is removed. The node shows as researched when the panel is reopened and there's now no button to purchase.

Tested by building the client and ran two local KSP instances against a local carreer mode server server.
Client 2's open panel now closes when the research hits. The science is deducted once and the node can't be re-purchased.

### DIFF
--- a/LmpClient/Systems/ShareTechnology/ShareTechnologyMessageHandler.cs
+++ b/LmpClient/Systems/ShareTechnology/ShareTechnologyMessageHandler.cs
@@ -34,11 +34,28 @@ namespace LmpClient.Systems.ShareTechnology
             System.StartIgnoringEvents();
             var node = AssetBase.RnDTechTree.GetTreeTechs().ToList().Find(n => n.techID == tech.Id);
 
+            //Check (before we respawn the tree below) whether the R&D screen currently has this exact
+            //node's detail panel open. If it does, its Research button will need to be refreshed afterwards.
+            var panelWasShowingThisNode = RDController.Instance && RDController.Instance.node_selected != null
+                && RDController.Instance.node_selected.tech != null
+                && RDController.Instance.node_selected.tech.techID == tech.Id;
+
             //Unlock the technology
             ResearchAndDevelopment.Instance.UnlockProtoTechNode(node);
 
             //Refresh the tech tree
             ResearchAndDevelopment.RefreshTechTreeUI();
+
+            //RefreshTechTreeUI() respawns the tree nodes but does NOT refresh an already open node panel.
+            //If another player had this node open, its Research button stayed active and let them buy the
+            //node again, spending the science a second time (see issue #667). We can't just call UpdatePanel()
+            //because the respawn orphans node_selected (its cached RDTech.state is never re-synced), so instead
+            //we close the stale panel. The node correctly shows as researched once it is reopened.
+            if (panelWasShowingThisNode && RDController.Instance)
+            {
+                RDController.Instance.node_selected = null;
+                RDController.Instance.ShowNothingPanel();
+            }
 
             //Refresh the part list in case we are in the VAB/SPH
             if (EditorPartList.Instance) EditorPartList.Instance.Refresh();


BraveCaperCat2 (approved):
Looks good. Is it not possible to re-open the panel after we've closed it, though?

GrahamKowalski:
My thinking was the close tells the other players "something has happened". But to be honest I didn't look at reopening the node.

ShiralynDev:
Would be cool if it could reopen the panel.

GrahamKowalski:
Good idea, I think it's better as a follow-up.

LmpClient/Systems/ShareTechnology/ShareTechnologyMessageHandler.cs
```diff
@@ -34,12 +34,29 @@ private static void TechnologyResearch(TechNodeInfo tech)
...
+            //RefreshTechTreeUI() respawns the tree nodes but does NOT refresh an already open node panel.
+            //If another player had this node open, its Research button stayed active and let them buy the
+            //node again, spending the science a second time (see issue #667). We can't just call UpdatePanel()
+            //because the respawn orphans node_selected (its cached RDTech.state is never re-synced), so instead
+            //we close the stale panel. The node correctly shows as researched once it is reopened.
+            if (panelWasShowingThisNode && RDController.Instance)
+            {
+                RDController.Instance.node_selected = null;
+                RDController.Instance.ShowNothingPanel();
+            }
```
BraveCaperCat2, line 58:
```suggestion
            if (panelWasShowingThisNode && RDController.Instance)
            {
                var selected_node = RDController.Instance.node_selected;
                RDController.Instance.node_selected = null;
                RDController.Instance.ShowNothingPanel();
                RDController.Instance.node_selected = selected_node;
                RDController.Instance.ShowNodePanel(selected_node);
            }
```
There might be a better way to do this, (there is a `RDController.Instance.UpdatePanel()` function, not sure if that would work or not) but as long as it fixes the issue, it shouldn't matter.

ShiralynDev:
I thought this closed the entire window but it doesent. Looks great.

ShiralynDev (approved):
LGTM!

PR #713, opened 2026-06-25
Fix development environment (on the client)

BraveCaperCat2:
### Fixes included in this PR:
- Fix the C# Dev Kit extension for VSCode not loading the client project.
### Changes proposed in this PR:
- Update the `LmpClient/LmpClient.csproj` file to use the new format already used elsewhere.
- Add the documentation for `Assembly-CSharp.dll`.
- Comment out some stuff in `LmpClient/Properties/TypeForwarding.cs‎` that was erroring.
- Some new files added from me building the mod.

PR #726, opened 2026-07-21
Fix LunaCompat TypeLoadException by compiling LmpClient type forwarders

DrewBanyai:
## Summary

Fixes a `System.TypeLoadException` that prevented mods compiled against the old, merged `LmpClient` (such as `LunaCompat`) from loading. The mods reference `LmpCommon` types (e.g. `LmpCommon.Message.Base.FactoryBase`) as living in `LmpClient`, and those references failed to resolve at runtime.

## Root cause

`LmpClient/Properties/TypeForwarding.cs` contained the `TypeForwardedTo` entries needed for mod compatibility, but the file was **never added to `LmpClient.csproj`**, so it was never compiled. As a result, none of its forwarders (`FactoryBase`, `MessageBase<>`, `MessageStore`, `MessageData`, `CliMsgBase<>`, the message interfaces, `ClientState`, `LogLevels`, `LmpVersioning`, `Common`, the `*CliMsg` types, etc.) were emitted into `LmpClient.dll`.

Only the forwarders in `AssemblyInfo.cs` were compiled in, and that set did **not** include `FactoryBase`. When `LunaCompat` was loaded, the runtime could not resolve `FactoryBase` (nor the other missing types) via `LmpClient`, throwing:

```
System.TypeLoadException: ... Could not resolve type with token 01000032
(from typeref, class/assembly LmpCommon.Message.Base.FactoryBase, LmpClient ...)
```

Simply adding `TypeForwarding.cs` to the build would have broken compilation, because ~20 forwarders were duplicated between `TypeForwarding.cs` and `AssemblyInfo.cs`, which raises compiler error CS0739 (duplicate `TypeForwardedTo`).

## Changes

- **`LmpClient/Properties/TypeForwarding.cs`** — reorganized into the single, deduplicated **superset** of every forwarder previously spread across both files (nothing that used to be forwarded is lost; the previously-dead `FactoryBase`, interfaces, `MessageStore`, etc. are now actually emitted).
- **`LmpClient/Properties/AssemblyInfo.cs`** — removed all `TypeForwardedTo` entries (and the now-unused `using`), leaving only assembly metadata so the forwarders have a single home.
- **`LmpClient/LmpClient.csproj`** — added `<Compile Include="Properties\TypeForwarding.cs" />` so the forwarders are actually compiled into `LmpClient.dll`.

No runtime logic changes; this only affects assembly metadata (type forwarders) emitted into `LmpClient.dll`. The forward targets live in `LmpCommon.dll`, which is already deployed alongside `LmpClient.dll`.

## Test plan

- [x] `LmpClient` builds cleanly in Debug (0 errors, no CS0739).
- [x] `LmpClient` builds cleanly in Release (0 errors, no CS0739).
- [x] Confirmed `TypeForwardedTo` now exists in exactly one compiled file, eliminating duplicate-forwarder collisions.
- [x] Confirmed `LmpCommon.dll` (the forward target) is produced in the build output and deployed with `LmpClient.dll`.
- [x] Load a mod that references the forwarded types (e.g. `LunaCompat`) in KSP and verify it loads without a `TypeLoadException`.

### DIFF
--- a/LmpClient/LmpClient.csproj
+++ b/LmpClient/LmpClient.csproj
@@ -477,6 +477,7 @@
     <Compile Include="Network\NetworkMain.cs" />
     <Compile Include="Network\NetworkStatistics.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
+    <Compile Include="Properties\TypeForwarding.cs" />
     <Compile Include="Systems\Chat\ChatMessageHandler.cs" />
     <Compile Include="Systems\Chat\ChatMessageSender.cs" />
     <Compile Include="Systems\Chat\ChatSystem.cs" />
--- a/LmpClient/Properties/AssemblyInfo.cs
+++ b/LmpClient/Properties/AssemblyInfo.cs
@@ -1,6 +1,5 @@
 ﻿using System.Reflection;
 using System.Runtime.InteropServices;
-using System.Runtime.CompilerServices;
 
 [assembly: AssemblyTitle("Luna Multiplayer Mod")]
 [assembly: AssemblyDescription("Luna Multiplayer Mod (client)")]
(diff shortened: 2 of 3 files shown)


LmpClient/LmpClient.csproj
```diff
@@ -477,6 +477,7 @@
     <Compile Include="Network\NetworkMain.cs" />
     <Compile Include="Network\NetworkStatistics.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
+    <Compile Include="Properties\TypeForwarding.cs" />
```
BraveCaperCat2, line 480: Not needed once #713 is merged.

BraveCaperCat2 (approved):
Approved.

LmpClient/LmpClient.csproj
DrewBanyai: Resolve the merge conflicts with it when you get the chance and I'll review asap

TheXankriegor:
This is for fixing loading of LunaCompat in 0.30.0 master builds, correct? I worry a little that making LunaCompat not immediately break with that version could cause bug reports due to some odd differences between the versions down the line. "Officially supporting" both versions at once would be an issue.
I don't know the differences to 0.29.2 in detail, so if that can be ruled out its all good 👍

ShiralynDev (approved):
LGTM

BraveCaperCat2:
> This is for fixing loading of LunaCompat in 0.30.0 master builds, correct? I worry a little that making LunaCompat not immediately break with that version could cause bug reports due to some odd differences between the versions down the line. "Officially supporting" both versions at once would be an issue. I don't know the differences to 0.29.2 in detail, so if that can be ruled out its all good 👍

Yes, but it only fixes the LunaCompat mod's issues. The server plugin would have to be updated to use .NET 10 (among other things) before it would be compattible.

PR #729, opened 2026-08-21
Remove LAN bypass in IntroduceToServer to fix false-positive connection failures

t1garbiznisbrate-ship-it:
### Summary of Changes
Removes the `ServerIsInLocalLan` bypass from `IntroduceToServer` in favor of standard Lidgren NAT introduction for all server connections.

### Problem & Background
Previously, `NetworkServerList.IntroduceToServer` checked `ServerIsInLocalLan(serverInfo.ExternalEndpoint)` to determine if a server was on the same local network. If true, it skipped the Master Server NAT punch and attempted a direct connection to the server's internal endpoint.

However:
1. **IPv4 CGNAT / Shared IPs:** On Carrier-Grade NAT (common on mobile data, fiber ISPs, universities, and VPNs), multiple distinct players share the exact same public IPv4 address. LMP falsely marked them as "local", skipped NAT punch, and tried to connect to a private internal IP (`192.168.x.x`), which failed completely.
2. **IPv6 Subnet Issues:** Attempting to infer LAN locality on IPv6 via subnet prefixes has similar edge cases (e.g., shared VPS subnets, stateful IPv6 firewalls requiring hole punching).

### Solution
- Removed the `ServerIsInLocalLan` method and its check in `NetworkServerList.IntroduceToServer`.
- All connections now proceed through the Master Server's NAT introduction flow (`MsIntroductionMsgData`).
- **Why this works seamlessly:** In Lidgren, `peer.Introduce` already sends punch packets to **both** the internal and external endpoints in parallel:
  - On genuine LAN/localhost connections, the internal packet arrives almost instantaneously (< 1ms) and connects locally.
  - On Internet / CGNAT connections, the external NAT punch succeeds without false-positive LAN skips.

### Verification
- [x] Ran `dotnet test`: all 35 unit tests passed with 0 errors.
- [x] Tested local dedicated server connection in KSP: connected via Master Server introduction with 0 delay.
- [x] Tested remote public server connection in KSP: successfully punched NAT and connected to public WAN server.

### DIFF
--- a/LmpClient/Network/NetworkServerList.cs
+++ b/LmpClient/Network/NetworkServerList.cs
@@ -1,4 +1,4 @@
-﻿using Lidgren.Network;
+using Lidgren.Network;
 using LmpClient.Systems.Ping;
 using LmpCommon;
 using LmpCommon.Message.Data.MasterServer;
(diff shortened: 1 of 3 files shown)


BraveCaperCat2:
We already have issues with clients being regarded as "local" when they aren't on the same network with IPv4. I think it's safe to leave this optimization missing, at least for now.

t1garbiznisbrate-ship-it:
@BraveCaperCat2 

> We already have issues with clients being regarded as "local" when they aren't on the same network with IPv4. I think it's safe to leave this optimization missing, at least for now.


Looking deeper into it, Lidgren's NAT introduction already sends punch packets to both the internal and external endpoints in parallel anyway (which connects almost instantaneously on actual LANs). The ServerIsInLocalLan bypass seems to do more harm than good due to IPv4 CGNAT / shared public IPs falsely bypassing NAT punch.

I have updated the PR to remove the ServerIsInLocalLan check entirely so all connections (IPv4 and IPv6) use standard Master Server NAT introduction. I tested this locally in-game against both local and public WAN servers, and it resolves the CGNAT false positives while keeping LAN/localhost connections working.

For extra context on why removing `ServerIsInLocalLan` is completely safe:

When looking at how `Lidgren.Network` handles NAT introductions (`Lidgren/NetNatIntroduction.cs` in `HandleNatIntroduction`):

```csharp
// Send internal punch
m_unsentUnconnectedMessages.Enqueue(new NetTuple<NetEndPoint, NetOutgoingMessage>(remoteInternal, punch));

// Send external punch
m_unsentUnconnectedMessages.Enqueue(new NetTuple<NetEndPoint, NetOutgoingMessage>(remoteExternal, punch));
```

Whenever the Master Server introduces a client and host, Lidgren **simultaneously transmits punch packets to both the internal LAN endpoint AND the external WAN endpoint**:

* **On actual LAN / localhost:** The packet targeting `remoteInternal` arrives almost instantaneously (<1ms) with zero firewall friction, and `HandleNatIntroductionSuccess` immediately receives the local endpoint to connect.
* **On the Internet / CGNAT:** The `remoteInternal` packet is silently dropped if unreachable, while the `remoteExternal` packet punches NAT hole state and establishes the connection across the WAN.

In the early days of LMP (`56ad9a41`), `ServerIsInLocalLan` was added as a manual shortcut to bypass the master server when `externalIP == ownIP`. Back then, residential IPv4 addresses were mostly unique per household. But with modern Carrier-Grade NAT (CGNAT on 4G/5G, fiber, Starlink, campus networks), distinct players frequently share the same public IPv4, causing `ServerIsInLocalLan` to trigger false positives and try connecting to unreachable RFC1918 private IPs.

Because Lidgren already provides built-in dual-endpoint punching, the manual LAN bypass in `LmpClient/Network/NetworkServerList.cs` was redundant and removing it eliminates the false positives while keeping local LAN play working natively.

PR #730, opened 2026-08-25
Fix Breaking Ground deployed science synchronization

Decappi:
> **AI-assisted contribution note**
>
> Most of the implementation in this PR was produced with AI assistance rather than written by me directly.
>
> I reproduced the original bug, worked through the investigation, reviewed the resulting diff, and iterated on it through multiple rounds of code review and testing. I also tested the final patch extensively in-game.
>
> I understand the problem being fixed and the intended behavior of the changes, but I do not claim to fully understand or be able to defend every implementation detail. Please review the implementation accordingly. I can provide reproduction steps, logs, save/server state, and test requested changes against the original bug.

# The annoying bug

I was running into a Breaking Ground deployed-science synchronization problem in LMP.

Deployed science worked correctly when initially placed, but its deployment-derived bonuses could be lost after synchronization or reconnect.

For example:

- an experiment deployed by a Scientist could initially have the expected deployment bonus, but later fall back to the base science rate;
- a solar panel deployed by an Engineer could initially produce its boosted power output, but later lose it;
- the ground-science controller could report stale or incomplete experiment/power state.

Picking up and redeploying an affected part fixed it temporarily, until the bad state was synchronized again.

## Reproduction

The original problem could be reproduced with a Breaking Ground deployed-science setup roughly like this:

1. Deploy a ground-science controller and experiment using a Scientist, so the experiment receives a deployment bonus.
2. Deploy the required power source using an Engineer.
3. Verify that the experiment and solar panel initially show the expected boosted values.
4. Let LMP synchronize the vessels/scenario.
5. Disconnect and reconnect to the server, or restart KSP and connect again.
6. Inspect the deployed experiment, power source and controller.

Before this patch, on the affected setup I could end up with the experiment falling back to its base deployment value and/or the deployed solar output being lost.

The incorrect state could then become part of the server-side `DeployedScience` state, making it persist across subsequent reconnects.

The investigation showed that this was not one isolated assignment but a combination of several lifecycle/order problems around deployed vessels, `persistentId` ownership and the `DeployedScience` scenario.

With the final patch applied, I can no longer reproduce the original deployed-science state loss using this flow.

# Fixes included in this PR

## 1. Defer transient deployed-part vessel snapshots

Breaking Ground parts pass through a transient deployment state where the proto snapshot is not yet suitable to become authoritative state.

If an outgoing vessel proto contains a ground module where `beingDeployed` is still true while `deployedOnGround` or `isEnabled` has not settled yet, LMP now defers that send.

Once deployment has settled, the vessel is serialized again from a fresh snapshot and synchronization continues normally.

This prevents a mid-deployment vessel definition from becoming canonical on the server.

## 2. Preserve `persistentId` ownership transactionally during vessel replacement

Incoming vessel replacement can interact badly with KSP's global persistent-part-ID registries.

The patch now treats temporary registry changes as a transaction:

- stale owners of the incoming wire `persistentId` values are evicted before construction/load;
- genuine foreign owners are preserved so stock collision handling can still do its job;
- removed registrations remain recoverable until the vessel replacement has completely succeeded;
- the transaction is committed only after final successful load/swap;
- validation, construction and post-load failures roll it back;
- rollback only restores owners that are still genuinely present in the live vessel set;
- registrations created for a failed incoming vessel are cleaned up.

This also avoids treating a replacement vessel with a reused vessel GUID as if it were the same object as the old owner.

## 3. Apply `DeployedScience` only after the initial vessel sync is complete

The stock `DeployedScience` scenario can be loaded before all corresponding deployed-science vessels exist locally.

When that happens, stock may interpret legitimate controller/experiment entries as missing.

The patch therefore defers authoritative `DeployedScience` scenario application until initial vessel synchronization is complete.

A readiness gate compares the membership declared by the incoming scenario with the deployed-science parts actually present in the live vessel set.

Outgoing `DeployedScience` synchronization remains blocked until the authoritative scenario has been successfully applied.

## 4. Relink already-loaded runtime modules after authoritative scenario load

By the time the deferred scenario is applied, some deployed-science `PartModule` instances already exist and may still contain stale/default runtime state.

After authoritative scenario application, the patch reconciles those live modules with the scenario:

- stale `scienceClusterData` caches are invalidated;
- stock cluster power/UI notifications are refreshed;
- `ScienceModifierRate`, `ScienceLimit` and `ScienceValue` are restored from the authoritative scenario into the corresponding live experiment modules;
- persisted solar power inputs are restored;
- `ActualPowerUnitsProduced` is recomputed using stock-equivalent line-of-sight and timewarp semantics.

The relinker does not overwrite accumulated science progress fields.

## 5. Recover deployed-science parts that missed live registration

There is also a lifecycle case where a deployed-science vessel exists correctly in the live game but never becomes registered in the stock deployed-science cluster dictionary.

A periodic 2-second sweep checks for these missing live registrations and replays the stock registration path.

The operation is idempotent.

The sweep reads stock's private `beingDeployed` field through reflection. If the field cannot be resolved, it fails closed and treats the part as still deploying rather than risking registration of transient state.

# Safety / scope

- No automatic deletion or migration of historical/stale clusters.
- Runtime relinking does not overwrite accumulated science progress fields.
- Genuine foreign `persistentId` owners are preserved.
- Failed vessel replacements restore only still-valid previous owners.
- Live deployed-science registration is idempotent.
- Uncertain deployment state fails closed rather than being synchronized.

# Validation

Automated validation:

- all five commits build independently;
- final Release build: **0 errors**;
- tests: **82/82 passed**;
- `git diff --check`: clean;
- replaying the five patches with `git am` onto a fresh `master` reproduces the expected final source tree;

In-game regression testing included:

- cold KSP start + server connect;
- warm LMP disconnect/reconnect;
- repeated reconnects;
- multiple scenario synchronization cycles;
- Scientist deployment bonus preservation;
- Engineer deployed-solar output preservation;
- server-side `DeployedScience` state preservation;
- normal vessel loading with no observed regression.

### DIFF
--- a/LmpClient/LmpClient.csproj
+++ b/LmpClient/LmpClient.csproj
@@ -458,6 +458,9 @@
     <Compile Include="VesselUtilities\OwnVesselReloader.cs" />
     <Compile Include="VesselUtilities\VesselSerializer.cs" />
     <Compile Include="VesselUtilities\TransientDeployedStateDetector.cs" />
+    <Compile Include="VesselUtilities\PersistentIdEvictionPolicy.cs" />
+    <Compile Include="VesselUtilities\VesselLoadTransaction.cs" />
+    <Compile Include="VesselUtilities\StalePersistentIdEvictor.cs" />
     <Compile Include="Systems\VesselSwitcherSys\VesselSwitcherSystem.cs" />
     <Compile Include="Windows\Admin\AdminDrawer.cs" />
     <Compile Include="Windows\Admin\AdminWindow.cs" />
--- a/LmpClient/LmpClient.csproj
+++ b/LmpClient/LmpClient.csproj
@@ -457,6 +457,7 @@
     <Compile Include="VesselUtilities\DiscoveryInfoSanitizer.cs" />
     <Compile Include="VesselUtilities\OwnVesselReloader.cs" />
     <Compile Include="VesselUtilities\VesselSerializer.cs" />
+    <Compile Include="VesselUtilities\TransientDeployedStateDetector.cs" />
     <Compile Include="Systems\VesselSwitcherSys\VesselSwitcherSystem.cs" />
     <Compile Include="Windows\Admin\AdminDrawer.cs" />
     <Compile Include="Windows\Admin\AdminWindow.cs" />
@@ -510,6 +511,7 @@
     <Compile Include="VesselUtilities\DockingPortUtil.cs" />
     <Compile Include="Systems\VesselProtoSys\VesselProtoMessageHandler.cs" />
     <Compile Include="Systems\VesselProtoSys\VesselProtoMessageSender.cs" />
+    <Compile Include="Systems\VesselProtoSys\TransientVesselSendHandler.cs" />
     <Compile Include="Systems\VesselProtoSys\VesselProtoSystem.cs" />
     <Compile Include="Systems\VesselRemoveSys\VesselRemoveEvents.cs" />
     <Compile Include="Systems\VesselRemoveSys\VesselRemoveMessageHandler.cs" />
--- a/LmpClient/LmpClient.csproj
+++ b/LmpClient/LmpClient.csproj
@@ -553,6 +553,7 @@
     <Compile Include="Systems\ModApi\ModApiMessageHandler.cs" />
     <Compile Include="Systems\Mod\ModFileHandler.cs" />
     <Compile Include="Systems\Scenario\ScenarioStructures.cs" />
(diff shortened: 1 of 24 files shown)


ShiralynDev:
I'm against the huge amount of comments in the code. I might be alone in this but the code feels like a second focus compared to comments.

Decappi:
> I'm against the huge amount of comments in the code. I might be alone in this but the code feels like a second focus compared to comments.

Yep, fair point, fixed. I generally prefer code to document itself too.

PR #737, opened 2026-09-09
Add "spectate" button to allow spectating players directly

Matszwe02:

This PR adds "spectate" button in the player list, allowing to spectate players without the need of going into tracking station and searching for the proper vehicle.


LmpClient/Windows/Status/StatusWindow.cs
ShiralynDev, line 1: Why a file mode change?

Matszwe02: Maybe I missed that when copying code between PCs

Matszwe02: Fixed, thanks for noticing

BraveCaperCat2: Apparently this happens when a file is, for example, made executable (linux thing?)
I'm not sure why you'd want to make a `.cs` file executable though so it's probably something else that triggers the same behaviour

BraveCaperCat2 (approved):
Looks good.

PR #739, opened 2026-09-09
feat: add time offset display during warping

Matszwe02:

When warping, players can still see how much time behind/ahead is given player, no need to stop warping to know that.


BraveCaperCat2 (approved):
Looks good.

